PR #224, opened 2026-05-10
refactor(rooms): room-centric heating refactor

asnowfix:
## Summary

Room-centric heating refactor across 6 phases. Makes **room** the first-class unit: each room owns its temperature sensors, door/window sensors, heaters, and an iCal-based occupancy agenda. Heater scripts run an autonomous six-gate control loop that survives daemon and internet loss for at least one day.

Also fixes #200: the README mistakenly labelled port 8888 as "Temperature control"; it is actually the Gen1 HTTP→MQTT proxy. Room/temperature management was always on port 6080 via MQTT RPC.

### Phases completed

- **Phase 0** — `myhome/temperature/` renamed to `myhome/rooms/`; DB tables renamed; in-memory caches removed; all reads hit SQLite directly
- **Phase 1** — `myhome/electricity/` package: `FixedWindowPricer`, publishes `myhome/electricity/status` every 15 min
- **Phase 2** — `myhome/weather/` package: Open-Meteo fetcher, 4-slot distiller, publishes `myhome/weather/forecast`; persists for offline fallback
- **Phase 3** — `myhome/ical/` package: iCal URL fetcher, VEVENT parser, publishes `myhome/rooms/<id>/agenda` daily; `room.seticurl` / `room.geticurl` RPC methods
- **Phase 4** — `myhome/devices/impl/setup.go`: `SetupRoom` classifies devices by role (heater / temp sensor / door sensor / motion sensor), derives MQTT topics, pushes to heater KVS; daemon runs at startup + daily; `myhome ctl room setup [room-id]` CLI
- **Phase 5** — `heater.js` split into two scripts within Shelly resource limits:
  - `heater-data-relay.js` (4 MQTT subs, 1 timer) — bridges daemon MQTT publications to device KVS for offline resilience
  - `heater-controller.js` (2–3 MQTT subs, 2 timers) — six-gate control loop; reads all external policy from KVS; frost override timer
- **Phase 6** — Enhanced Rooms tab:
  - Room cards: kind badges, setpoints, iCal indicator, today's agenda (busy/free + next transition), assigned devices by role
  - Room edit modal: iCal URL field, loaded and saved via `room.edit`
- **Phase 7** — `setup.go` uses `shelly.IsBluDevice()` for BLU detection (covers all typed prefixes); `MacFromShellyID()` for topic derivation; motion sensors classified but not wired to heater gates
- **Phase 8** — `myhome/electricity/`: `MultiIntervalPricer` replaces `FixedWindowPricer`; `--cheap-electricity=HH:MM-HH:MM[,HH:MM-HH:MM]` flag; example: `23:15-07:15,12:00-14:00`
- **Phase 9** — Docs & issues: electricity section updated; filed #235 (solar pricer) and #236 (ENEDIS Tempo pricer)
- **Phase 10** — Weather location auto-discovery: `pkg/shelly/cloud` package (`Cloud.GetStatus`); `Shelly.DetectLocation` response type registered; daemon orchestrates priority chain: config → Shelly cloud → IP geolocation (ip-api.com)
- **Phase 11** — Housekeeping: `docs/rooms-refactor.md` removed; `heater-data-relay.js` filename comment fixed

### Six heater gate conditions

| # | Condition | Source |
|---|---|---|
| 0 | Internal temp ≤ 7 °C (frost override, bypasses 1–5) | local MQTT sensor |
| 1 | Electricity cheap now | KVS `room/electricity` (heater-data-relay) |
| 2 | Upcoming weather slot below eco setpoint | KVS `room/weather` (heater-data-relay) |
| 3 | Home globally occupied (within 12 h) | KVS `room/occupancy` (heater-data-relay) |
| 4 | Room agenda says occupied now (iCal busy slot) | KVS `room/agenda` (heater-data-relay) |
| 5 | All doors/windows closed | local MQTT sensors |

## Test plan

- [x] `make test` passes (all Go packages)
- [ ] Upload `heater-data-relay.js` + `heater-controller.js` to a real heater device and verify gate logging
- [ ] Verify `heater.js` still works until new scripts are deployed
- [ ] Manually verify Rooms tab in browser: cards show kinds, setpoints, devices; edit modal saves iCal URL

## Commits

- docs(rooms): add refactor plan for room-centric heating ([f05e3bd](https://github.com/asnowfix/home-automation/commit/f05e3bd1728f42a586836ce59d9c6bfe3bbdb5f5))
- refactor(rooms): phase 0 — rename temperature→rooms, DB-first, no cache ([6d053ac](https://github.com/asnowfix/home-automation/commit/6d053ac722d6800d087d96649117d7c38bbefac4))
- feat(electricity): phase 1 — electricity pricing interface & MQTT publisher ([fec3ee3](https://github.com/asnowfix/home-automation/commit/fec3ee387ebb694728290fa816a1426bc146a59f))
- feat(weather): phase 2 — weather proxy via Open-Meteo ([4ccc95c](https://github.com/asnowfix/home-automation/commit/4ccc95c9664762b6f5a15be4a77bba6f6abf0248))
- feat(ical): phase 3 — per-room iCal agenda fetcher & MQTT publisher ([cf3c2af](https://github.com/asnowfix/home-automation/commit/cf3c2af5901e56de7ebc18f8a5c5722811abfc97))
- feat(rooms): phase 4 — automatic heater device setup ([006ed69](https://github.com/asnowfix/home-automation/commit/006ed6995154394c9fc0389013a1c0167557194a))
- feat(heater): phase 5 — split heater.js into data-relay + heater-controller ([27393b1](https://github.com/asnowfix/home-automation/commit/27393b1ab66f1cb649d0d012052186fdc30bbc60))
- refactor(heater): rename data-relay.js → heater-data-relay.js ([b744bb3](https://github.com/asnowfix/home-automation/commit/b744bb39a712ccda5c5951e2be5ea6dd7f667724))
- feat(ui): phase 6 — enhanced Rooms tab ([107e4a7](https://github.com/asnowfix/home-automation/commit/107e4a70bf062e800020d84f53a20d72531f83f2))
- merge: resolve conflicts with main (events feature) ([712eb86](https://github.com/asnowfix/home-automation/commit/712eb8677ed58f1e1e23344ee7b0eaabaab776ff))
- docs(readme): fix port 8888 label — Gen1 proxy, not temperature control ([b7097f4](https://github.com/asnowfix/home-automation/commit/b7097f420cbb268552130b784cbcc44db6c95330))
- merge: resolve settings.json conflict with main ([cb752b1](https://github.com/asnowfix/home-automation/commit/cb752b19b2e9669fca3481eb82cd2cbfcc1acfa1))
- fix(rooms): use IsBluDevice() for BLU device classification in setup ([e6cc65e](https://github.com/asnowfix/home-automation/commit/e6cc65ecff32b0e426492aafa930332818d2bc0c))
- feat(electricity): multi-interval pricing; --cheap-electricity flag ([55bccb0](https://github.com/asnowfix/home-automation/commit/55bccb0c4b6187c57d85c19a61834dddb61ce3a8))
- feat(weather): location auto-discovery via Shelly cloud + IP geolocation ([d789cd8](https://github.com/asnowfix/home-automation/commit/d789cd8b305eaa99b8eee159923555b5bcb55700))
- Merge remote-tracking branch 'origin/main' into worktree-feature+temperature-refactor ([de4a03d](https://github.com/asnowfix/home-automation/commit/de4a03d972fc72de8efe380bd1d906045cf38369))
- Merge remote-tracking branch 'origin/main' into worktree-feature+temperature-refactor ([351d1e7](https://github.com/asnowfix/home-automation/commit/351d1e7a7032dd23dbdc77ffaa0bb9e86d2586d5))
- test(daemon): cover shellyLocation branches, raise coverage floor to 22% ([0323886](https://github.com/asnowfix/home-automation/commit/03238864628da7e7072396039cdb6a0b1eb94ab1))


docs/configuration.md
```diff
@@ -516,3 +522,81 @@ Or use flag:
...
+Topic: `myhome/electricity/status` (retained, QoS 1)
+
+```json
+{"cheap": true, "until_epoch": 1234567890}
+```
+
+- `cheap`: `true` if the current time is within the cheap window
+- `until_epoch`: Unix timestamp when the current period (cheap or expensive) ends
+
+## Weather Forecast Configuration
```
asnowfix, line 593: The weather forecast configuration should have an "shelly" mode (the default) by which it calls to an available shelly device that has cloud connectivity one (via https://shelly-api-docs.shelly.cloud/gen2/ComponentsAndServices/Cloud#cloudgetstatus-example) and gets the curent latitude & longitude (via https://shelly-api-docs.shelly.cloud/gen2/ComponentsAndServices/Shelly#response-6).

You can also look into online geolocation database & suggest other providers in addition to "shelly"

docs/rooms-refactor.md
```diff
@@ -0,0 +1,334 @@
+# PLAN: Room-Centric Heating Refactor
```
asnowfix: Drop that plan from the PR, when fully implemented (included comments taken into account)

docs/rooms-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+| `office` | 08:00–18:30 | — |
+| `kitchen` | 06:30–09:00, 18:00–21:00 | 09:00–11:00, 18:00–21:00 |
+| `other` | — | — |
+
+`—` means eco mode always; frost override (gate 0) still active.
+All values are configurable per room; these are just the seeded defaults.
+
+---
+
+## Sensor MQTT topic derivation
```
asnowfix: There are now additional (or replacement) topic names depending on BLU device types (motion, humidity & temperature, door/window opening):  adapt the plan & implementation accordingly.

docs/rooms-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+- Rename DB tables via migrations (SQLite `ALTER TABLE RENAME`)
+- Add `ical_url` column to `rooms`
+- Add `weather_cache` and `room_agenda_cache` tables
+- Remove `Service.rooms`, `Service.kindSchedules`, `Service.weekdayDefaults` maps
+- All `Get*` methods read SQLite directly; `Set*` methods write and return
+- Update RPC method names to `room.*` namespace (audit: some may already be correct)
+- Update `internal/myhome/temperature.go` types to match new table names
+- `make test` must pass
+
+## Phase 1 — Electricity pricing interface `[DONE]`
```
asnowfix: Document the electricity pricing interface & the available implementation.  Research what could be used to provide an implementation for ENEDIS (the electricity provider in France: static configuration or live api).

internal/shelly/scripts/heater-data-relay.js
```diff
@@ -0,0 +1,155 @@
+// data-relay.js
```
asnowfix: script name has changed

myhome/ctl/options/options.go
```diff
@@ -68,6 +68,11 @@ var Flags struct {
 	EnableGen1Proxy             bool
 	EnableOccupancyService      bool
 	EnableTemperatureService    bool
+	EnableElectricityService    bool
+	ElectricityCheapStart       string  // HH:MM, e.g. "23:15"
+	ElectricityCheapEnd         string  // HH:MM, e.g. "07:15"
```
asnowfix: There can be multiple cheap electricity slots per day, per provider (eg. when solar centrals provide more, the home electricity provider may statically plan to set chap interval in the afternoon), only one of them potentially crossing midnight (expanding to the next day).  The cheap-electricity should be an array of intervals.

by default, there is just one provider (ENEDIS in France for me at the moment), but my solar panels could also be such a provider, whose cheap interval may overlap on the main provider. Keep this possibility of an additional provider a placeholder & log a follow-up issue to implement it later.

internal/shelly/scripts/heater-controller.js
```diff
@@ -0,0 +1,435 @@
...
+    if (!data || !data.slots || data.slots.length === 0) {
+      cb("gate-2-weather-unavailable");
+      return;
+    }
+    var h = new Date().getHours();
+    var slots = data.slots;
+    var horizon = CONFIG.cheapHorizonH;
+    var threshold = CONFIG.setpoints.eco;
+    for (var i = 0; i < slots.length; i++) {
+      if (slots[i].h >= h && slots[i].h <= h + horizon && slots[i].t < threshold) {
```
asnowfix, line 201: **Midnight-window bug.** `slots[i].h >= h && slots[i].h <= h + horizon` never matches next-day hours: at `h = 23` with `horizon = 2`, a slot for 00:00 or 01:00 has `h = 0/1` and fails `>= 23`. The default cheap-electricity window is 23:15–07:15, so the weather gate mis-evaluates exactly during the main heating window. The daemon publishes slots in chronological order — walk the first `horizon+1` entries of `data.slots` instead of comparing hour-of-day, or publish epoch timestamps per slot.

myhome/rooms/setup_rpc.go
```diff
@@ -0,0 +1,24 @@
...
+// RegisterSetupHandler registers the room.setup RPC method with the given backend.
+func RegisterSetupHandler(handler RoomSetupHandler) {
+	myhome.RegisterMethodHandler(myhome.RoomSetup, func(ctx context.Context, params any) (any, error) {
+		p := params.(*myhome.RoomSetupParams)
+		if p.RoomID != "" {
+			return handler(ctx, p.RoomID)
+		}
+		// Empty room_id = setup all rooms; return a placeholder result.
+		// The daemon handles this via SetupAllRooms directly.
+		return &myhome.RoomSetupResult{}, nil
```
asnowfix, line 22: **Silent no-op contract break.** `myhome ctl room setup` with no argument reaches this branch and returns an empty placeholder — nothing calls `SetupAllRooms`, but the CLI help promises "If room-id is omitted, all rooms with assigned devices are set up." Either pass a second handler for the all-rooms case (the daemon already has `SetupAllRooms`) or make the CLI require the argument. Also: `params.(*myhome.RoomSetupParams)` two lines up is an unchecked assertion — use the checked `v, ok :=` form like the EventList handler (this is the same panic pattern #357 just removed elsewhere).

myhome/weather/weather.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil, fmt.Errorf("longitude %g out of range [-180, 180]", lon)
+	}
+
+	f := &Forecaster{
+		lat:     lat,
+		lon:     lon,
+		log:     log.WithName("weather.Forecaster"),
+		mc:      mc,
+		db:      db,
+		httpGet: http.Get,
```
asnowfix, line 77: **No timeout on external fetch** (CLAUDE.md internet-optional rule: "Always add a timeout and a fallback"). `http.Get` uses `http.DefaultClient`, which has no timeout: a blackholed connection blocks `fetchAndPublish` forever inside `Run`'s loop, so the ticker never fires again and the cache fallback is unreachable. Same issue in `ical.go:53`. Inject `(&http.Client{Timeout: 15 * time.Second}).Get` here instead of `http.Get` — keeps the test seam.

myhome/weather/location.go
```diff
@@ -0,0 +1,40 @@
...
+type ipAPIResponse struct {
+	Status string  `json:"status"`
+	Lat    float64 `json:"lat"`
+	Lon    float64 `json:"lon"`
+}
+
+// IPGeoLocation fetches coordinates from ip-api.com (free, no authentication).
+func IPGeoLocation(ctx context.Context) (float64, float64, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet,
+		"https://ip-api.com/json?fields=status,lat,lon", nil)
```
asnowfix, line 19: **Dead code as written — verified today**: `curl https://ip-api.com/json?fields=status,lat,lon` returns `{"status":"fail"}` — ip-api.com's free tier is HTTP-only (HTTPS requires a paid key). The `http://` variant returns coordinates. So the geolocation fallback never succeeds, and any installation without a cloud-connected Shelly and without configured lat/lon silently gets no weather (which then blocks heater gate 2). Use `http://` with a comment about the tradeoff, or switch providers. Also give this call its own `context.WithTimeout` — the caller passes the daemon-lifetime ctx, which never expires.

myhome/rooms/storage.go
```diff
@@ -0,0 +1,421 @@
...
+		{"temperature_rooms", "rooms"},
+		{"temperature_kind_schedules", "room_kind_schedules"},
+		{"temperature_weekday_defaults", "room_weekday_defaults"},
+	}
+
+	for _, r := range renames {
+		var count int
+		s.db.QueryRow(
+			"SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name=?", r.old,
+		).Scan(&count)
```
asnowfix, line 80: **Data-loss path on ignored error.** If this probe `Scan` fails (locked DB, etc.), `count` stays 0, the `temperature_rooms → rooms` rename is silently skipped, and `createTables` then creates a fresh empty `rooms` table — existing room config vanishes from the app while still sitting in the legacy table. Check the `QueryRow(...).Scan` error and abort the migration on failure (same for the `pragma_table_info` probe below; the deliberate ignore on the `ALTER TABLE ... ADD COLUMN` needs an explicit `_, _ =` + comment or it will fail errcheck once this branch rebases onto main's new lint gate). Also consider the `...Context` query variants throughout this file — #364 will sweep exactly this and new code shouldn't add to that backlog.

myhome/devices/impl/setup.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		return roleUnknown
+	}
+
+	if strings.HasPrefix(id, "shellyht-") {
+		return roleTempSensor
+	}
+
+	// Non-BLU, non-HT → heater candidate
+	return roleHeater
```
asnowfix, line 50: **Over-broad heater classification.** Every non-BLU, non-`shellyht-` device in a room becomes a "heater": the daily `SetupAllRooms` then writes `room-id` + `script/heater/*` KVS keys to every Gen2 plug/switch/pool/garden controller that has a room assigned. Devices have limited KVS slots, and other scripts load KVS with `match:"*"`. Gate this on evidence the device actually runs the heater script (Script.List via RPC, cached) or on an explicit role/tag, rather than "not a sensor". Related: after pushing a changed temperature/door topic, `heater-controller` reloads config but never re-subscribes — first-time setup leaves the script subscribed to nothing until a manual restart. Have setup restart the script when topic keys changed, or implement resubscription in the script.

asnowfix (commented):
## Proposed-changes review of DRAFT PR #224 (not an approval; PR stays owner-driven)

Reviewed at 0323886 against merge-base abe0b86, with the 2026-07-18 boundary decisions in mind (#388: single `myhome.db`, temperature/rooms storage owned by this PR; #368 campaign waves #359/#362/#364/#366 pending). Architecture-level findings first; inline comments mark the exact spots for the sharpest bugs.

### A. Interaction with the #368 campaign — sequencing recommendation

1. **Merge order**: recommend landing #224 (after the fixes below) **before** wave #359 (module collapse) and #366 (layout renames). This PR adds three new sub-modules (`myhome/electricity`, `myhome/ical`, `myhome/weather`) plus the renamed `myhome/rooms` module — consistent with the *current* per-package-module pattern, but every one of them is a `go.mod` #359 will delete. If #359 lands first, this PR conflicts in `go.work` + every new `go.mod`; if #224 lands first, #359 just collapses 3 more modules mechanically. The `temperature→rooms` rename before #366's `myhome/* → internal/` move is also the right order (rename once, then move once).
2. **`daemon.go` is the #1 conflict hotspot.** `daemon.Run` gained ~100 lines of wiring here, while the campaign has #362 (composition root) and #364 (errgroup restructure of `Run`) queued against the same function, and the integration branch already rewrote `NewDaemon` (PR #380). Strong recommendation: extract all new wiring (`discoverLocation`, electricity/weather/ical/rooms startup, the daily `SetupAllRooms` ticker goroutine) into a separate file, e.g. `myhome/daemon/rooms_services.go` with one `startRoomServices(...)` entry point, so `Run`'s diff footprint is a few lines and survives the campaign's restructures.
3. **Make new code born campaign-compliant** so the waves don't have to sweep it: (a) `rooms/storage.go`, `weather`, `ical` use non-context `Exec/Get/Select` — #364 sweeps exactly this; new code should use the `...Context` variants from day one. (b) `setup_rpc.go` has an unchecked `params.(*myhome.RoomSetupParams)` assertion — the same panic-prone pattern #357 just removed; use the checked form like the EventList handler does. (c) The unchecked `Scan`/`Exec` returns in `rooms/storage.go` will fail errcheck once #355's golangci-lint baseline (already on main's CI) sees this branch — rebase will surface them; fix pre-emptively. (d) Give the publishers `Run(ctx) error` signatures instead of `Run(ctx)` so #364's errgroup can adopt them unchanged.
4. **DB boundary vs #388: respected.** Rooms tables live in the shared `myhome.db` via `storage.DB()` — consistent with the single-DB decision, and #388 explicitly excludes these tables. One coordination point: when #364's shared open/pragma/migrate bootstrap helper lands, `rooms.NewStorage` should adopt it. Also, `weather_cache` and `room_agenda_cache` schemas are defined **twice** (in `rooms/storage.go:createTables` *and* in `weather.go`/`ical.go` `createTable`) — pick one owner per table or the copies will drift.
5. This branch is 106 commits behind main; rebase/merge main before further phases — among other things main's CI now runs golangci-lint/vet/-race on PRs.

### B. Resilience (binding CLAUDE.md rules)

6. **No HTTP timeouts on any external fetch.** `weather.fetch` and `ical.fetchAndParse` use bare `http.Get` (no timeout, no context); `IPGeoLocation` uses `http.DefaultClient` (context-carried but the caller passes the daemon-lifetime ctx, which never expires). A blackholed connection wedges the fetch loop forever — the cache fallback is never reached again. CLAUDE.md: "Always add a timeout and a fallback/no-op path before shipping any code that calls an external URL." Use a `http.Client{Timeout: ~15s}` (also fixes testability: the injectable `httpGet` can stay).
7. **`discoverLocation` runs synchronously inside `daemon.Run`, before `mc.Start()` and `ui.Start`.** It iterates every Gen2 device with HTTP RPCs, then calls ip-api.com. Combined with (6), a hung external call stalls the entire daemon startup — UI never comes up. That violates internet-optional (remote sources must not block local operation). Move the whole location+weather startup into a goroutine.
8. **The IP-geolocation fallback is dead code as written** — verified empirically today: `https://ip-api.com/json` returns `{"status":"fail"}` (free tier is HTTP-only; `http://` works and returns coordinates). As shipped, any home without a cloud-connected Shelly and without configured lat/lon silently never gets weather. Switch to `http://` (coarse-location tradeoff worth a comment) or a different provider.
9. **The "survives daemon+internet loss for at least one day" claim doesn't match the gate logic.** Gate 3 blocks on occupancy `ts` older than **12 h**, so heating stops ~12 h into a daemon outage (frost override excepted). Meanwhile gate 1 ignores the published `until_epoch` entirely (`data.cheap !== true` only), so a daemon outage *freezes* the cheap flag — if it died during the cheap window, the device considers electricity cheap indefinitely (heats at peak rates). And gate 4 applies cached slots with no date/staleness check (yesterday's agenda replays today). Please decide and document one coherent degraded-mode policy per gate (CLAUDE.md requires the degraded mode in the PR description), and make gate 1 honor `until_epoch` with a fallback to the pricer schedule shape.
10. **Default-closed gates make heating impossible on a fresh/partial install.** A room with no iCal URL never gets an agenda published → `gate-4-agenda-empty` blocks forever; no occupancy service → gate 3 blocks. So a heater in a room without a configured calendar can never heat except frost mode. Owner decision needed: which gates default **open** when their data source is absent/never-published (vs. absent-because-stale)? My suggestion: missing key = open (feature not configured), stale key = per-gate policy.

### C. Correctness bugs

11. **Gate 2 midnight-window bug** (inline comment): `slots[i].h >= h && slots[i].h <= h + horizon` can never match next-day hours (h=23 → slots with h∈{0,1} fail `>= 23`), and hour slots for "tomorrow morning" duplicate hours-of-day. Since the default cheap window is 23:15–07:15, this misfires exactly in the primary operating hours. Compare in slot-sequence space (slots are already time-ordered) rather than hour-of-day space, and consider publishing epoch timestamps per slot instead of bare hour numbers.
12. **`myhome ctl room setup` (no args) is a silent no-op** (inline comment): the RPC handler returns an empty placeholder for empty `RoomID` and nothing invokes `SetupAllRooms`, but the CLI help promises "all rooms with assigned devices are set up".
13. **Every non-BLU, non-HT device is classified as a heater** (inline comment): daily `SetupAllRooms` then writes `room-id` + `script/heater/*` KVS keys to every Gen2 device assigned to any room — plugs, pool Pro3, garden controller. Devices have limited KVS slots and other scripts read KVS with `match:"*"`. Gate heater classification on actual heater-script presence (Script.List) or an explicit role/tag.
14. **Pushed topics never take effect without a manual script restart.** `heater-controller`'s `kvs_rev` handler deliberately reloads only non-subscription config, so a new `internal-temperature-topic`/`door-sensor-topics` from `room setup` (first setup, or sensor replacement) leaves the old subscriptions in place — on first install the script subscribes to nothing and never controls. Either resubscribe on change (MQTT.unsubscribe exists in the Shelly API) or have the daemon restart the script via RPC after a KVS push that changed topics.
15. `rooms/storage.go` migration: the `sqlite_master`/`pragma_table_info` probe `Scan` errors are discarded — if a probe fails, `count` stays 0, the rename is silently skipped, and `createTables` creates a fresh empty `rooms` table: existing room config silently disappears while the old `temperature_rooms` table still holds it. Check the errors (inline comment).

### D. Shelly JS (mostly good)

The scripts follow the repo's Espruino rules well: functions defined before use, `var`-only, non-empty catches with `e` referenced, `"prop" in obj`, manual trimming, and `runGates` is a textbook sequential-async pattern that avoids nesting depth. Budget is within limits (relay 4 subs/1 timer; controller ≤3 subs/2 timers). Two notes:

16. `checkStaleness` fires 4 concurrent `KVS.Get` calls; a same-moment MQTT burst adds concurrent `KVS.Set`s — that flirts with the 5-concurrent-RPC limit. Use the AGENTS.md single-timer task-queue pattern to serialize (same reason serial KVS writes are required on the Pro3).
17. The agenda payload carries no date, so the device can't detect day rollover; and the daily refresh is anchored to daemon start time, not midnight — start the daemon at 18:00 and each day's agenda arrives at 18:00, meaning the device runs on yesterday's slots from 00:00–18:00 (compounding finding 9's gate-4 gap). Include the date in the payload and align the refresh to local midnight. Related: the iCal parser ignores RRULE, so *recurring* events (the common case for room agendas) only match their first occurrence date — implement or document loudly.

### E. Housekeeping

18. `.claude/settings.json` gains blanket `Write`/`Edit` and broad Bash allows inside a feature PR — split into its own PR or confirm it's deliberate; it changes the security posture for every future agent session on this repo.
19. Electricity flag help says "auto-enabled with device manager" and run.go does implement it — good; make sure `docs/configuration.md`/`myhome-example.yaml` cover all five new options (4-file rule).

### Suggested priority for the next iteration
Blockers-if-this-were-final: 6, 7, 8, 9/10 (policy), 11, 12, 13, 15. Sequencing: 1–4. The rest can trail.

asnowfix:
## Triage 2026-09-06 — **maintainer decision needed**; the only one of the eleven whose subject is untouched by recent work

Reviewed while clearing the `stale-unreviewed` backlog. 54 files, +3833/−981, 18 commits, **332 commits
behind `main`**, **3 conflict regions** — notably *fewer* than several much smaller PRs here, because
heating is the one area campaign #401 never touched.

**This is the strongest rescue candidate among the large PRs**, and the judgement is genuinely the
maintainer's rather than mine, for two reasons:

1. **It is a design commitment, not a bug fix.** Making *room* the first-class unit — owning its
   sensors, door/window contacts, heaters and an iCal occupancy agenda — is an architectural direction.
   No amount of rebasing decides whether that is the wanted model; only you can.
2. **Its stated resilience properties match the project's own rules.** The six-gate heater control loop
   surviving daemon and internet loss for at least a day is exactly the "internet-optional" and
   "daemon-optional per device" requirements in `CLAUDE.md`. That is a point in its favour, and it also
   means the device-side scripts need real hardware verification before merge — on heater devices, none
   of which is a current test bed.

**One incidental fix should be salvaged regardless of the PR's fate**: it corrects the README's claim
that port 8888 is "Temperature control" (it is the Gen1 HTTP→MQTT proxy; room/temperature management is
on 6080 via MQTT RPC), closing #200. That is a two-line change trapped behind a 4.8 kLOC refactor. If
this PR is closed or parked, lift that fix out separately so #200 does not stay open for a documentation
error nobody disputes.

**Recommendation:** decide the architecture question first, in the issue, before anyone spends time
rebasing. If the room-centric model is wanted, this branch is worth reviving in phases (it is already
organised into six). If it is not, close it and lift the README fix.

asnowfix:
-

PR #225, opened 2026-05-10
fix(sfr,proxy): retry SFR gateway discovery; add --remote-proxy for SSH remote access

asnowfix:
## Summary

- **SFR nil gateway crash**: `getBoxIp()` used `sync.Once`, so a failed discovery at startup meant `boxIp` stayed `nil` permanently. Every subsequent call passed `nil` to `queryBox`, producing the URL `http://<nil>/api/1.0/` and crashing with _"no such host"_. Removed `sync.Once`; the mutex-guarded path now retries on every call until a gateway responds. Added nil guard in `GetHostsList`/`GetDnsHostList` for a clean error message. Added `SFR_BOX_IP` env var to pin the address when gateway auto-discovery fails (multi-homed hosts, VPN, remote access).

- **Remote proxy for SSH access**: the reverse proxy at `localhost:6080` TCP-dials device IPs directly. When working remotely over SSH this always fails — the local machine can't reach `192.168.1.1`. Added `--remote-proxy <url>` flag (also `daemon.remote_proxy` config key / `MYHOME_DAEMON_REMOTE_PROXY` env) that forwards all `/devices/...` requests to a remote myhome daemon instead of dialing devices directly.

## Test plan

- [ ] Verify `myhome run` still reaches the SFR box normally when on the home network
- [ ] Verify that when the SFR box is unreachable at startup, the error log now says _"SFR box IP not discovered (set SFR_BOX_IP env var to override)"_ instead of _"lookup \<nil\>: no such host"_, and the daemon continues to retry every minute
- [ ] Set `SFR_BOX_IP=192.168.1.1` and confirm it is used without running gateway discovery
- [ ] With `ssh -L 6081:localhost:6080 home-pi`, run `myhome run --remote-proxy http://localhost:6081` locally and confirm `http://localhost:6080/devices/192.168.1.1/` is reachable

## Commits

- fix(sfr,proxy): retry SFR gateway discovery; add --remote-proxy for SSH remote access ([1ff3e1a](https://github.com/asnowfix/home-automation/commit/1ff3e1a5b1a7dcac0d76a3c57f47412295f809f8))
- fix(mqtt): use net.JoinHostPort for IPv6-safe broker URL construction ([3f03bfe](https://github.com/asnowfix/home-automation/commit/3f03bfe03b1e6e42eb79e549e9978243ac2a2c6d))
- fix(net): use net.JoinHostPort everywhere host:port is assembled ([7a0c502](https://github.com/asnowfix/home-automation/commit/7a0c5025e0bc88b933375f74b475d8eacd9c86b6))


asnowfix:
Added a second commit (3f03bfe) fixing the IPv6 broker URL bug:

`fmt.Sprintf("%s:%d", ip, port)` produces `::1:1883` for IPv6 addresses — Go's net package then rejects it with `too many colons in address`. All four host:port assembly sites in `lookupBroker` (and the `"me"` branch) are now using `net.JoinHostPort`, which correctly produces `[::1]:1883`.

PR #227, opened 2026-05-14
fix(events): repair event log UI — storage, dedup, naming, BLU types

asnowfix:
## Summary

- **Events DB path**: default to `events.db` (relative, same dir as `myhome.db`) — eliminates `SQLITE_CANTOPEN` caused by `~/.myhome/` never being created; `os.MkdirAll` guard added for explicit absolute paths
- **Gen2 listener**: handle `NotifyStatus` for switch on/off events (previously only `NotifyEvent` was processed); populate `data` with `apower`/`voltage` for metered switches; fix BLU timestamp/deduplication by flooring gateway `ts` to integer seconds
- **Event log UI**: resolve device names via `GetDeviceByAny` + MAC-from-Shelly-ID fallback; fix embedded-struct name collision that printed raw Go structs in the Event column; device filter now resolves names/MACs to actual `device_id` values via `IN` clause
- **BLU device IDs**: infer type-specific prefix from event capabilities (`shellybludoorwindow2`, `shellybluht3`, `shellyblumotion1`, `shellyblubutton1`); opportunistic upgrade of existing `shellyblu-<mac>` entries on next event via new `DeviceRegistry.RenameDevice`
- **`pkg/shelly.MacFromShellyID`**: extract MAC from Shelly device ID suffix — used in `NewDeviceFromMqttId` (no network call needed) and in the UI resolver

## Test plan

- [ ] Start daemon with `go run ./myhome daemon run -B <broker>` — confirm `events.db` is created in the working directory alongside `myhome.db`, no `SQLITE_CANTOPEN` error
- [ ] Toggle a switch via the UI — confirm `switch.on` / `switch.off` rows appear in `/event-log` within seconds
- [ ] Check that metered switches (Shelly 1 Mini G3) show `{"apower":…,"voltage":…}` in the Data column
- [ ] Trigger a BLU motion sensor — confirm single row (not 3 duplicates), correct timestamp, device name shown
- [ ] Type a device name in the event log filter — confirm rows filter correctly
- [ ] Confirm existing `shellyblu-<mac>` devices are renamed to typed IDs on next BLU event
- [ ] `make test` passes

## Commits

- fix(events): repair event log UI — storage, dedup, naming, BLU types ([8badbc7](https://github.com/asnowfix/home-automation/commit/8badbc7f7c2903d1452f774c399f19759eb0760e))
- fix(zeroconf): prevent panic on browser restart after network loss ([b601070](https://github.com/asnowfix/home-automation/commit/b601070d88aaeaf55a352e0d9c818ed3ba1cc0d1))

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -79,6 +79,7 @@ Ports: 6080 (dev web UI), 80 (systemd), 6060 (pprof), 9100 (Prometheus).
 - **Config options**: adding any new option requires updating 4 files — `options.go`, `run.go`, `docs/configuration.md`, `myhome-example.yaml`. Env var pattern: `MYHOME_<SECTION>_<KEY>`.
 - **RPC handler tests**: tests that call `myhome.RegisterMethodHandler()` must restore state in `t.Cleanup` and must not call `t.Parallel()` (shared package-level map).
 - **Database migrations**: Use `COUNT(*)` (returns int) not bool when checking SQLite column existence. See AGENTS.md "Database Patterns".
+- **SQLite database paths**: new databases use a plain relative filename (e.g. `"foo.db"`), matching `myhome.db`. Do not invent a new default directory (e.g. `~/.myhome/`, XDG paths) unless all existing databases already use it. If a flag or config key lets the user supply an absolute path, the `NewStorage` constructor must call `os.MkdirAll(filepath.Dir(path), 0o755)` before opening the file — SQLite cannot create missing parent directories.
 - **File moves**: always `git mv`, never delete-and-recreate (preserves `git log --follow` history).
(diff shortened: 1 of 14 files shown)


asnowfix:
Added a second commit (b601070) that fixes the crash observed after network loss.

**Root cause**: `grandcat/zeroconf` closes the entries channel inside `LookupParams.done()` when `Browse` exits. The `scan` channel was created once outside the restart loop and passed to every `BrowseService` call. After the first browse session ended (triggered by the network outage), zeroconf closed `scan`. On the next restart, the same closed channel was passed to a new `Browse` call; when zeroconf tried to close it again → `panic: close of closed channel`.

**Fix**: move `scan := make(chan *zeroconf.ServiceEntry, 1)` inside the `for` loop so each Browse call gets a fresh, unclosed channel.

PR #259, opened 2026-06-13
chore(learning): apply modernize fixes (interface{} → any, strings.Builder)

asnowfix:
## Summary

Learning exercise — Go modernization patterns applied mechanically via the `modernize` static analyzer:

- **`cmd/datacollector/main.go`** — 24× `interface{}` → `any` (struct fields, function signatures, map literals); brings this file in line with the 361 existing `any` usages elsewhere in the repo
- **`myhome/metrics/exporter.go`** — replace `string +=` in loop with `strings.Builder` to avoid O(n²) allocations in `GetAll()`

## How

```bash
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix ./...
```

No behavioural changes. Rebased onto latest `main`; `make build` and `make test` both pass.

## Context

Part of a self-paced Go learning plan — exercise A1 (generics & stdlib modernization). This is a **learning branch, not intended to merge** — kept as a draft for reference/review.

### DIFF
--- a/cmd/datacollector/main.go
+++ b/cmd/datacollector/main.go
@@ -27,16 +27,16 @@ import (
 
 // APICall represents a single API call with request and response
 type APICall struct {
-	Timestamp   time.Time   `json:"timestamp"`
-	DeviceID    string      `json:"device_id"`
-	DeviceName  string      `json:"device_name"`
-	DeviceModel string      `json:"device_model,omitempty"`
-	Method      string      `json:"method"`
-	Channel     string      `json:"channel"`
-	Request     interface{} `json:"request"`
-	Response    interface{} `json:"response"`
-	Error       string      `json:"error,omitempty"`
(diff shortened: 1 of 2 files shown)


asnowfix:
## Triage 2026-09-06 — **no action; deliberately left as-is**

Reviewed while clearing the `stale-unreviewed` backlog, and listed here only so nobody re-derives its
status on the next pass.

This is a **learning-plan branch** and is **draft by design**. It is not abandoned work awaiting a
merge/close decision like the other ten, and it should not be swept up in one. Modest scope (2 files,
+35/−35 — `interface{}` → `any`, `strings.Builder`), and nothing about it has decayed in a way that
needs fixing.

Left untouched: not rebased, not closed, not re-reviewed. If it should be merged, that needs the
maintainer's explicit word; absent that, draft is the correct resting state.

PR #267, opened 2026-06-14
feat(minifier): set ES5 target and extract prometheus-metrics script

### LINKED ISSUE #266: fix(prometheus-metrics): minified script crashes Espruino with "Too much recursion"
## Summary

`prometheus-metrics.js` (extracted from `watchdog.js`) crashes immediately on start when uploaded with minification. The error from the device:

```
Uncaught Error: Too much recursion - the stack is about to overflow
 at ...ower_watts",this.TYPE_GAUGE,[n],"Instant power consumption i...
in function "_buildSwitchMetrics" called from ...s._buildSwitchMetrics(e,t,o),this._publishSwitchMetrics(e,t)...
in function "generateMetricsForSwitch" called from ....generateMetricsForSwitch(o);return n
```

Without minification (`--no-minify`) the script runs fine.

## Root Cause

`tdewolff/minify` aggressively merges sequential statements into comma expressions and pulls them into the `return`. A function with **two return paths** like:

```javascript
if (!sw) { return ""; }
// ... N assignments and calls ...
return result;
```

gets minified into:

```javascript
return sw ? (assign1, assign2, ..., callN, result) : ""
```

Espruino's expression evaluator is **recursive in C**: each element of the comma expression and each sub-expression adds one C stack frame. With 21 elements (6 `printPrometheusMetric` calls + 6 `publishIndividualMetric` calls + assignments) the C stack overflows before the function returns.

### Why restructuring to a single return is not enough

Even after restructuring `generateMetricsForSwitch` to use a single return (`var result=""; try{ if(sw){...} }catch(e){...} return result`) and extracting `_buildSwitchMetrics` / `_publishSwitchMetrics`, the minifier still collapses `_buildSwitchMetrics`'s 6 sequential `n += printPrometheusMetric(...)` statements into a 6-element comma expression **in the return**:

```javascript
return n="", n+=printPrometheusMetric(...), ..., n
```

When `printPrometheusMetric` is evaluated inside that comma chain, it enters *its own* 5-element comma-in-return. The combined evaluator depth across `generateMetricsForSwitches → generateMetricsForSwitch → _buildSwitchMetrics → printPrometheusMetric` still exceeds Espruino's stack limit.

## Minifier Fix (not yet merged)

Set `Version: 2009` on the `tdewolff` minifier (`pkg/shelly/script/main.go`):

```go
m.AddFunc("text/javascript", (&mjs.Minifier{Version: 2009}).Minify)
```

`minVersion(2015)` gates template literal output; `minVersion(2019)` gates optional catch binding. Setting `Version: 2009` disables both. It does **not** suppress the comma-expression-in-return optimisation, so this alone is not sufficient to fix the crash — but it prevents two other Espruino-incompatible constructs as a defensive measure.

## What Needs to Happen

The real fix requires ensuring no function in the metrics path produces a **deep comma expression anywhere** (not just in returns). Options to explore:

1. **Emit statements directly instead of building a string** — replace `result += metric(...)` chains with `MQTT.publish()` calls per metric, eliminating the string-accumulation pattern entirely.
2. **Break `_buildSwitchMetrics` into even smaller helpers** — one per metric — so each helper's minified body is a single expression, not a chain.
3. **Patch the minifier** — post-process the minified output to split long comma sequences at statement boundaries (hard to do reliably).
4. **Upload unminified** — acceptable for low-memory-pressure devices; the Prometheus script is ~9 KB minified, ~14 KB raw.

## Investigation Log

- Non-minified: `running: true`, `mem_used: 5712` ✓  
- Minified watchdog (all 3 modules): `errors: ["crashed"]` — root cause of the original bug
- Minified watchdog (without PrometheusMetrics): `running: true` ✓ — confirms PrometheusMetrics is the only crasher
- Minified prometheus-metrics.js (with `_buildSwitchMetrics` / `_publishSwitchMetrics` split): still crashes with same "too much recursion" in `_buildSwitchMetrics`

## Related

- Extracted in PR #??? (bugfix/watchdog-crash — removed PrometheusMetrics from watchdog.js to unblock the watchdog fix)
- Minifier `Version: 2009` change lives on branch `feature/minifier-es5-version` (draft PR, linked here)

asnowfix:
Fixes #266

## Summary

This PR holds two things extracted from the watchdog crash investigation (PR #268):

1. **`pkg/shelly/script/main.go`** — sets `Version: 2009` on the `tdewolff` JS minifier so it never emits ES6+ constructs (template literals, optional catch binding) that Espruino's modified JS engine does not support.

2. **`prometheus-metrics.js`** — the `PrometheusMetrics` side script, extracted from `watchdog.js` by PR #268 to unblock the watchdog fix.

## Status: draft — side script still crashes

`prometheus-metrics.js` crashes when minified with the same "Too much recursion" error documented in #266. The `Version: 2009` flag is necessary but not sufficient: the minifier still collapses sequential `result +=` chains into deep comma-expression-in-return patterns that overflow Espruino's C expression-evaluator stack. This PR stays draft until #266 is resolved.

## What needs to happen before merge

See #266 for the full investigation. Options:
- Emit metrics directly via `MQTT.publish()` per metric instead of building a string, eliminating the chain entirely.
- Break `_buildSwitchMetrics` into one helper per metric so each helper's minified body is a single expression.
- Upload `prometheus-metrics.js` unminified (`--no-minify`) as a short-term workaround.

## Test plan

- [ ] Resolve the deep-comma-expression crash in #266
- [ ] Upload `prometheus-metrics.js` minified to arrosage — `running: true`
- [ ] Verify metrics appear on MQTT topic `shelly/metrics/<device-id>`

### DIFF
--- a/internal/shelly/scripts/prometheus-metrics.js
+++ b/internal/shelly/scripts/prometheus-metrics.js
@@ -1,0 +1,286 @@
+// prometheus-metrics.js - Prometheus metrics exporter for Shelly devices
+// Publishes device metrics in Prometheus format via MQTT at a regular interval.
+
+var CONFIG = {
+    scriptName: "prometheus-metrics",
+    enableLogging: true,
+    prometheus: {
+        enabled: true,
+        publishIntervalSeconds: 30,
+        mqttTopic: "shelly/metrics",
+        publishIndividualMetrics: true
+    }
+};
+
+function log() {
+    if (!CONFIG.enableLogging) return;
+    var s = "";
+    for (var i = 0; i < arguments.length; i++) {
+        try {
+            var a = arguments[i];
+            if (typeof a === "object") {
+                s += JSON.stringify(a);
+            } else {
+                s += String(a);
+            }
+        } catch (e) {
+            s += String(arguments[i]);
+            if (e && false) {}
+        }
+        if (i + 1 < arguments.length) s += " ";
+    }
+    print(CONFIG.scriptName + ": " + s);
+}
+
+var PrometheusMetrics = {
+    // Constants
+    TYPE_GAUGE: "gauge",
+    TYPE_COUNTER: "counter",
+
+    // Device info
+    deviceInfo: null,
+    defaultLabelsStr: "",
+    metricPrefix: "shelly_",
+    emittedMeta: {},
+    publishTimer: null,
+
+    // Track switch states for activation/deactivation events
+    monitoredSwitches: [],
+    switchStates: {},
+    activationCounts: {},
+    deactivationCounts: {},
+
+    log: log.bind(this, "[PrometheusMetrics]"),
+
+    // Discover and initialize switch state tracking
+    initSwitchTracking: function() {
+        this.log("Discovering available switches...");
+
+        var availableSwitches = [];
+        for (var i = 0; i < 4; i++) {
+            var switchId = "switch:" + i;
+            var status = Shelly.getComponentStatus(switchId);
+            if (status && ("output" in status)) {
+                availableSwitches.push(switchId);
+                this.switchStates[switchId] = status.output || false;
+                this.activationCounts[switchId] = 0;
+                this.deactivationCounts[switchId] = 0;
+            }
+        }
+
+        this.monitoredSwitches = availableSwitches;
+        this.log("Found " + availableSwitches.length + " switch(es): " + JSON.stringify(availableSwitches));
+    },
+
+    // Check for switch state changes and track activation/deactivation
+    checkSwitchStateChanges: function() {
+        var list = this.monitoredSwitches;
+        for (var i = 0; i < list.length; i++) {
+            var switchId = list[i];
+            var sw = Shelly.getComponentStatus(switchId);
+            if (!sw) continue;
+
+            var currentState = sw.output || false;
+            var previousState = this.switchStates[switchId];
+
+            if (currentState !== previousState) {
+                if (currentState) {
+                    this.activationCounts[switchId]++;
+                    this.log("Switch " + switchId + " activated (count: " + this.activationCounts[switchId] + ")");
+                } else {
+                    this.deactivationCounts[switchId]++;
+                    this.log("Switch " + switchId + " deactivated (count: " + this.deactivationCounts[switchId] + ")");
+                }
+
+                this.switchStates[switchId] = currentState;
+
+                if (CONFIG.prometheus.publishIndividualMetrics) {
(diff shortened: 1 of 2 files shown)


asnowfix:
## Closing — both halves have landed independently, so nothing remains in this branch

Closed 2026-09-06 on the maintainer's instruction, while clearing the `stale-unreviewed` backlog. This
PR bundled two unrelated changes; **each reached `main` by its own route**, so there is nothing left
here to merge.

**1. `prometheus-metrics.js`** — already on `main` via **#268** (commit `9db5c97c`), which extracted it
from `watchdog.js` to unblock the watchdog fix. Landed independently of this PR.

**2. Setting an ES5 target on the JS minifier** — landed today via **PR #598** (`5de52281`), re-derived
against current `main` rather than cherry-picked, because the API had moved in the 370 commits since
this branch was written: today's vendored `github.com/tdewolff/minify/v2 v2.24.3` exposes `Version` as
a field on the `js.Minifier` struct, so the fix is `m.Add("text/javascript", &mjs.Minifier{Version:
2009})` rather than the call this branch used in June.

### What that change turned out to be worth

This PR's premise was right, and stronger than it claimed. `CLAUDE.md` lists **"never empty catch
blocks"** on its kill list — the rule that `catch (e) {}` becomes `catch {}` after minification and is a
syntax error on-device. That is ES2019 optional catch binding, and until #598 the minifier ran at its
**unbounded default target**, so the hazard was live and defended against only by asking every author
to remember the `if (e && false) {}` workaround.

The regression test added with #598 reproduces it from source for the first time — against `main`,
`catch (e) { fallback(); }` minified to `catch{fallback()}`. Blast radius was measured across all 19
scripts in `internal/shelly/scripts/`: **18 byte-identical, `prometheus-metrics.js` +4 bytes**, traced
to a `"\n"` that can no longer collapse to an ES2015 backtick string.

### What is explicitly NOT fixed

**#266 stays open.** This PR's own description was right that the ES5 flag is *"necessary but not
sufficient"* for `prometheus-metrics.js`'s "Too much recursion" crash — the minifier still collapses
`result +=` chains into deep comma expressions that overflow Espruino's C expression-evaluator stack.
The +4-byte observation from the survey is recorded on that issue rather than acted on.

**Neither change is verified on hardware.** #598 alters what would be uploaded to every device and has
not run on one. The risk direction is favourable — output is strictly more conservative and 18 of 19
scripts are unchanged byte-for-byte — but favourable is not measured. It is listed under "What is NOT
verified on hardware" in `docs/integration-tests.md`.

The branch `feature/minifier-es5-version` is left in place.

PR #287, opened 2026-06-25
fix(shelly): avoid panic when Foreach encounters a device with no ID yet

asnowfix:
## Summary

- `Foreach` now skips devices whose ID is empty (partially-discovered), logging at V(1) instead of crashing
- `NewDeviceFromSummary` returns an error for empty IDs instead of letting `UpdateId` panic
- New `pkg/shelly/device_test.go` covers both fixed paths plus `IsGen1Device`, `IsBluDevice`, `MacFromShellyID`, and `NewDeviceFromMqttId`
- Coverage floor raised 21 → 22

## Root cause

`go run ./myhome ctl shelly mqtt config '*' -B 192.168.1.2` panicked because the device list returned by `LookupDevices` can include entries whose ID has not been resolved yet. `Foreach` passed them straight to `NewDeviceFromSummary` → `UpdateId("")` → `panic("invalid device id: ")`.

## Test plan

- [ ] `make test` passes
- [ ] `./scripts/check-coverage.sh` reports ≥ 22%
- [ ] `go run ./myhome ctl shelly mqtt config '*' -B 192.168.1.2` no longer panics


asnowfix:
Closing in favour of a clean cherry-pick branch (#288 incoming) — original branch had too many unrelated commits ahead of v0.10.x.

PR #334, opened 2026-07-12
backport(#333): fix ldflags version import path

asnowfix:
## Summary
Backport of #333 to \`v0.11.x\`.

- \`myhome/Makefile\`'s \`GO_LDFLAGS\` used bare package names (\`-X version.Version=...\`) instead of the fully-qualified import path. The linker silently no-ops an \`-X\` flag that doesn't resolve, so \`Version\`/\`Commit\`/\`Program\`/\`Repo\` were never set by \`make build\`/\`make run\` or CI's \`build-pr-deb\` job — only tagged goreleaser builds (which already used the correct path) were unaffected.
- Symptom: "unknown" version shown in the CLI/UI on any packaged install without a \`.git\` directory — reproduced on the NAS after installing PR #332's artifact.

## Test plan
- [x] \`make build\`
- [x] \`make test\`
- [x] Copied the built binary outside the repo and ran \`myhome version\` — correctly printed the embedded version

## Commits

- fix(build): use fully-qualified import path for version ldflags ([3bbd565](https://github.com/asnowfix/home-automation/commit/3bbd5658643c2b684eab18831d02482d0d46fa0b))

### DIFF
--- a/myhome/Makefile
+++ b/myhome/Makefile
@@ -3,10 +3,12 @@
 #cmd
 PWD ?= $(shell cd)
 
-GO_LDFLAGS += -X version.Program=$(notdir $(PWD))
-GO_LDFLAGS += -X version.Repo=$(shell git config --get remote.$(shell git remote).url)
-GO_LDFLAGS += -X version.Version=$(shell git describe --tags --exact-match 2>/dev/null || git describe --always)$(shell git diff-index --quiet HEAD -- || echo '-dirty')
-GO_LDFLAGS += -X version.Commit=$(shell git rev-parse HEAD)
+VERSION_PKG := github.com/asnowfix/home-automation/pkg/version
+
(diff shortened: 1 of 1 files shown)


asnowfix:
The \`build\` check failure is a pre-existing issue on \`v0.11.x\`, not caused by this change: \`TestSendDigest\` in \`myhome/notice\` fails identically on stock \`origin/v0.11.x\` (verified locally before applying this commit). It's already fixed on \`main\` via commit 57d7911 ("fix(test): anchor TestSendDigest event timestamp to real time"), which is pending arrival on \`v0.11.x\` through PR #332.

This PR's actual change (\`myhome/Makefile\` ldflags path) builds and installs correctly — verified locally by running the built binary outside the repo and confirming the version is properly embedded rather than falling back to \`git describe\`.

Issue #368, opened 2026-07-14
umbrella: Go expert review — grade 11.5/20, findings & refactoring roadmap


asnowfix:
Umbrella for the full Go expert review of this repository performed 2026-07-13 against `main` @ 2cd3628. All findings are spread across the 13 self-contained issues listed below; this issue carries the overall judgement, the priority order, and the dependency graph, and stays open until the child issues are done.

## Overall grade: 11.5 / 20

Calibration: 10 = "works, but nervous to maintain"; 14 = "solid with localized debt"; 17+ = "exemplary open-source Go".

**Strengths (the hard-to-retrofit part — already here):**
- *Testing culture:* 74 test files, benchmarks, purpose-built fakes (`FakeDevice`, `RecordingMockClient`), CI coverage gate with ratcheted minimum + coverage-delta job on PRs. Best-in-class for a hobby project.
- *Storage layer:* WAL/synchronous pragmas with correct, well-reasoned comments; the `:memory:` connection-pool trap documented. Senior-level care.
- *Operational maturity:* systemd units/timers, Debian + MSI packaging, Prometheus, pprof, release automation, watchdogs, explicit resilience principles (internet-optional, daemon-optional).
- *Documentation habits:* ARCHITECTURE.md, AGENTS.md, plan files, comments that explain why.

**Weaknesses:**
- *Live correctness bugs in the RPC core* (#356): no request/response correlation under concurrency, a nil-channel deadlock path, methods bypassing their receiver for a global. This is the daemon's spine (≈ −3 points alone).
- *Global mutable state as default wiring* (#362): eight package-level singletons; transport libraries import the CLI flags package. The project documents rules to survive its globals ("don't call t.Parallel()") instead of removing them.
- *Zero static analysis* (#355): no vet, no lint, no `-race` in CI — and the mechanically-findable defects (duplicated profile block, ignored errors, dead assignments) are all still present.
- *Structural overhead & hygiene:* 59 Go modules for one binary (#359), duplicated MQTT abstraction (#363), committed goroutine/device dumps with real LAN IPs (#358), 40-line commented-out blocks, unidiomatic surface (`-E` suffixes, `Id_` fields, SCREAMING_SNAKE constants) (#366).

Trajectory: the weaknesses are concentrated and mechanical to fix — Wave 1+2 below lifts this to ~14/20; the strengths are the part you can't buy back later.

## Priority roadmap

**Wave 1 — guardrails first (start here):**
[...]

Issue #362, opened 2026-07-14
arch: dependency-injection pass — remove package-level singletons (TheClient, options.Flags in libraries, MQTT globals)


asnowfix:
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). Package-level mutable globals are the default wiring mechanism across the codebase. Inventory (each is state that functions reach for instead of receiving):

| Global | Where |
|---|---|
| `hlog.Logger` | `hlog/hlog.go:21` |
| `myhome.TheClient` | `internal/myhome/proxy.go:9` — used by all `myhome/ctl/*`, `Foreach`, `internal/myhome/blu`, `internal/myhome/config.go` |
| `theClient`, `mqttOps`, `mqttBroker` | `myhome/mqtt/client.go` |
| MQTT client registry (`SetClient`/`GetClient`, panics if set twice, `ResetClient()` for tests) | `pkg/shelly/mqtt/client.go:30-67` |
| `deviceMqttRegistry` | `pkg/shelly/device.go:110-114` |
| `options.Flags` + `options.ViperConfig` | `myhome/ctl/options/options.go` — read by **libraries**, not just the CLI |
| `signatures`/`methods` maps | `internal/myhome/methods.go` |
| `global.PanicOnBugs` | `internal/global` |

Observable costs, all documented in the repo itself: RPC-handler tests "must not call `t.Parallel()`" (CLAUDE.md), `ResetClient()` test-only hooks, "panic: MQTT client already set", and the bug where `client` methods call `TheClient` instead of their receiver.

**Worst single edge:** `myhome/mqtt` (the MQTT transport) imports `myhome/ctl/options` (the cobra/viper CLI flags package) for its configuration — `myhome/mqtt/client.go:16`. `myhome/daemon/daemon.go:17` reads flags the same way. A transport library must receive a config struct, not read CLI globals.

Related anti-pattern to fix in the same pass: `myhome/mqtt.client` stores a `context.Context` in a struct field (`myhome/mqtt/client.go` — `ctx context.Context // Process-wide context`); pass ctx per-method, keep only a cancel/close for shutdown.

## Proposed change

Make `myhome/daemon` (and `myhome/ctl`'s command setup) the **composition root**:

1. Constructors take their dependencies: `NewDaemon(log, mqttClient, storage, deviceManager, cfg)`, `NewClient(log, mc, timeout)`, etc. The MQTT client takes a plain `Config` struct built from flags at the edge.
[...]

Issue #364, opened 2026-07-14
robustness: goroutine lifecycle (errgroup), HTTP server timeouts, daemon startup error propagation, context-aware DB queries


asnowfix:
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). Non-test code has ~36 raw `go func` sites, one `sync.WaitGroup`, and zero `errgroup` uses — background goroutines (RPC server loop, MQTT watchdog, subscribers, proxies) have no join point, and shutdown is "cancel ctx and hope". Several concrete robustness defects come with that:

1. **Daemon startup failures are invisible.** `daemon.Start()` runs `go d.Run()` and discards `Run`'s error (`myhome/daemon/daemon.go:74-77`). If startup fails (DB locked, port taken), the service manager believes the daemon is up.
2. **Errors from ListenAndServe dropped.** `go http.ListenAndServe(...)` at `pkg/shelly/gen1/proxy.go:36` discards the error entirely — if the port is taken, the Gen1 proxy is silently absent forever. `myhome/daemon/daemon.go:98` (pprof server) logs it but neither restarts nor stops.
3. **HTTP servers without timeouts.** `&http.Server{Addr: addr}` at `internal/myhome/ui/server.go:36`; bare `ListenAndServe` for pprof. Set at least `ReadHeaderTimeout` (LAN-only softens but doesn't remove the concern).
4. **`os.Exit(0)` mid-command** at `myhome/ctl/ctl.go:93` skips all defers, including MQTT graceful disconnect and CPU-profile stop. Return instead.
5. **DB queries are cancellation-deaf.** `myhome/storage` and `myhome/temperature` use non-context `Exec/Query/Get/Select` almost everywhere (28 non-ctx vs 5 ctx call sites). Switch to the `...Context` variants so daemon shutdown and timeouts can cancel in-flight queries.

## Proposed change

- Restructure `daemon.Run` around `golang.org/x/sync/errgroup`: each long-lived service is `g.Go(...)`; `g.Wait()` propagates the first failure; deliberate policy per service — fatal vs degraded-with-retry — instead of "log and continue".
- Report startup errors through the service API (run to a "started" barrier synchronously, or signal readiness/failure on a channel).
- Add `ReadHeaderTimeout` to every `http.Server`; give the pprof listener its own `http.Server` value.
- Replace the `os.Exit(0)` with a normal return.
- Mechanical `...Context` sweep in the storage layers.

## Acceptance criteria

[...]

Issue #363, opened 2026-07-14
mqtt: unify the two parallel MQTT client abstractions (pkg/shelly/mqtt vs myhome/mqtt)


asnowfix:
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). The repo contains **two parallel MQTT client abstractions**:

- `pkg/shelly/mqtt` — a `Client` interface + QoS constants + a global set-once registry (`SetClient`/`GetClient(ctx)`, panic on double-set, `ResetClient()` for tests).
- `myhome/mqtt` — a near-identical `Client` interface (superset: adds `DeviceServer()`, `Start()`, `IsConnected()`, `Close()`, `SubscribeWithTopic`), duplicate QoS constants, and the only real implementation (eclipse/paho wrapper with watchdog/reconnect logic), plus its own global singleton.

`pkg/shelly` code obtains the client via its registry global; the daemon/CLI construct the real client and call `pkg/shelly/mqtt.SetClient(...)` to bridge the two. `myhome/daemon/daemon.go:21-22` even imports the same `myhome/mqtt` package twice under two aliases (`mqttclient`, `mqttserver`), manufacturing the illusion of two components.

## Proposed change

1. Keep **one** small consumer-side interface, on the library side — `pkg/shelly/mqtt` is the natural home since `pkg/shelly` must stay standalone-exportable (owner decision: `pkg/shelly`, `pkg/sfr`, `pkg/beem` remain separate modules). Trim it to what `pkg/shelly` actually consumes.
2. Keep **one** implementation: `myhome/mqtt`'s paho wrapper (app side), declared to satisfy the library interface. App-only capabilities (`DeviceServer`, watchdog control) stay on the concrete type, not the interface.
3. Delete the duplicated QoS constants and the duplicate interface.
4. Replace both global registries with explicit injection (companion issue: dependency-injection pass): `pkg/shelly` entry points take the `mqtt.Client` as a parameter — no `GetClient(ctx)` that can return `nil` on cancelled context.
5. Deduplicate the double import alias in `daemon.go`.

## Acceptance criteria

- [ ] One `Client` interface definition and one set of QoS constants in the repo.
- [ ] No `SetClient`/`GetClient`-style global bridge between the packages.
- [ ] `pkg/shelly` compiles standalone against only its own `mqtt` interface package.

## Dependencies

- Depends on: dependency-injection pass (globals removal) — or at least its MQTT part.
[...]

PR #378, opened 2026-07-16
fix(rpc): response correlation, start() error propagation, receiver-vs-global (#356)

### LINKED ISSUE #356: rpc client: no request/response correlation under concurrency; start() failure deadlocks; methods bypass receiver for global
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). The MQTT RPC client in `internal/myhome/client.go` — the transport used by every `myhome ctl` command and by daemon-internal calls — has three related correctness bugs. These are live bugs, independent of any refactoring ambition.

## Bug 1 — No request/response correlation: concurrent calls cross-wire

`client.CallE` publishes a request carrying a random `request_id`, then waits for **whatever message arrives next** on its subscription channel (`internal/myhome/client.go:186`: `case resStr = <-hc.from:`). The response's `Dialog.Id` is never compared with the request's.

Failure scenario: two goroutines call `CallE` concurrently (the web UI issues concurrent HTMX requests; `Foreach` fans out). Each caller can receive the *other's* response — producing type-confused results or spurious unmarshal errors that are near-impossible to reproduce.

**Fix:** one dispatcher goroutine reads `hc.from`, unmarshals only the `Dialog` envelope, and routes the payload by `Dialog.Id` to a per-call channel (`map[string]chan response` guarded by a mutex, entry removed on delivery/timeout). This also makes timeouts naturally per-call.

## Bug 2 — `start()` swallows errors → send on nil channel blocks forever

`client.start()` returns nothing. If `Subscribe` or `Publisher` fails it only logs (`internal/myhome/client.go:50-62`) — but `hc.me` is already set, so subsequent calls skip re-initialization, and `hc.to` stays `nil`. The next `CallE` executes `hc.to <- reqStr`: **a send on a nil channel blocks forever**, ignoring the context.

**Fix:** `start(ctx) error`; on failure reset state so a retry is possible; `CallE` propagates the error. The send should also be wrapped in a `select` with `ctx.Done()`.

## Bug 3 — Methods on `client` call the global singleton instead of the receiver

`LookupDevices` and `ForgetDevices` are methods on `hc *client` but invoke `TheClient.CallE(...)` (`internal/myhome/client.go:104-106,133`) instead of `hc.CallE(...)`. This only works while exactly one client exists and the `TheClient` package global (`internal/myhome/proxy.go:9`) has been assigned. Any second client instance (tests, embedded use) silently calls through the wrong object or nil-panics.

**Fix:** use the receiver. (Full removal of the `TheClient` global is a separate dependency-injection issue; this fix is safe and local.)

## Additional hardening in the same file (optional, same PR welcome)

- `CallE` returns `Method{}` as the `any` result on the unknown-method path (`client.go:151`) — return `nil`.
- Timeout uses `time.After(hc.timeout)` alongside ctx — fold into `context.WithTimeout`.
- Replace `RandStringBytesMaskImprRandReaderUnsafe` with `crypto/rand` + `hex.EncodeToString` named `newRequestID`.

## Acceptance criteria

- [ ] A race-enabled test drives two concurrent `CallE` calls against a fake MQTT client with deliberately reordered responses; each caller gets its own response.
- [ ] A test makes `Subscribe` fail; `CallE` returns an error promptly instead of hanging.
- [ ] No method of `client` references `TheClient`.

## Dependencies

Best landed after the CI hardening issue (lint + `-race` jobs) so the new tests run under the race detector in CI, but not blocked by it.

**Issue links:** soft dependency on #355 (race job in CI).

asnowfix:
Closes #356. Part of umbrella #368 (Wave 1).

## What

Fixes the three live correctness bugs in the MQTT RPC client (`internal/myhome/client.go`):

1. **Request/response correlation under concurrency** — `CallE` used to read whatever arrived next off the shared response channel, cross-wiring concurrent callers. A single `dispatch` goroutine (sole reader of the subscribe channel) now routes each response to its caller through a `pending` map keyed by `Dialog.Id`, guarded by its own mutex. The response channel is registered *before* the request is published, so dispatch can never race ahead of registration; unmatched responses are logged and dropped.
2. **`start()` nil-channel deadlock** — a failed `Subscribe`/`Publisher` left `hc.me` set and `hc.to` nil, so every subsequent `CallE` blocked forever sending on a nil channel. `start()` now returns an error that `CallE` propagates, commits state only on success (so a later call retries), and the publish send is `select`-guarded by `ctx`.
3. **Receiver vs global** — `LookupDevices`/`ForgetDevices` called through the package-level `TheClient` singleton instead of their receiver; redundant `start()` pre-calls removed (`CallE` starts lazily).

Also replaces the bespoke `RandStringBytesMaskImprRandReaderUnsafe` helper (`random.go` deleted) with a straightforward crypto/rand + hex request-id generator.

## Tests

New race-enabled whitebox tests (`internal/myhome/client_test.go`) with a deterministic fake transport (blocking channel handoffs, no `time.Sleep`, no polling):
- `TestCallE_ConcurrentCalls_CorrelateResponsesByDialogID` — two concurrent calls, responses deliberately fed back in reversed order.
- `TestCallE_SubscribeFailure_ReturnsErrorPromptly` — the old deadlock path now returns an error.
- `TestCallE_ContextCancelDuringSend_ReturnsPromptly` — canceled ctx unblocks an undrained publish channel.
- `TestLookupDevices_UsesReceiverNotGlobalTheClient` — poisons `TheClient` (restored via `t.Cleanup`, no `t.Parallel`) and drives an independent instance.

## Verification

- `go build ./...` OK
- `make test`: rc=0 (46 package runs ok, 0 FAIL)
- `golangci-lint run ./internal/myhome/...`: 0 issues
- `go test -race -count=1 ./internal/myhome/...`: all pass (including the 4 new tests)

## Notes

- Implemented by the issue agent; finalized (3 staticcheck QF1008 fixes, verification, commit) by the coordinator after repeated API-overload cutoffs.
- `TheClient` global itself is intentionally left in place — its removal is #362's scope.

## Commits

- fix(rpc): correlate responses by Dialog.Id, propagate start() errors, drop TheClient global usage ([dceab4a](https://github.com/asnowfix/home-automation/commit/dceab4a35a89b8c70283b8563be3ce979a6df022))
- fix(rpc): bound the publish send by hc.timeout as well ([353d156](https://github.com/asnowfix/home-automation/commit/353d15684036d766c7789fd927e228c9571485b2))


asnowfix (commented):
# Review — issue #356 (RPC client correctness) — verdict: APPROVE

Reviewed the full diff and the final file state at dceab4a (correctly based on integration tip 25bf682, post-#355), with special attention to the coordinator-finalized bits.

## Verification performed (scratch worktree)

- `go build ./...` OK; `golangci-lint run ./internal/myhome/...` (pinned v2.5.0): 0 issues.
- `go test -race -count=2 ./internal/myhome/`: pass; new tests stressed with `-race -count=20`: pass, deterministic (no sleeps/polls, blocking channel handoffs as claimed).
- Grepped the whole repo: `client.go:145` was the **only** caller of `RandStringBytesMaskImprRandReaderUnsafe` — `random.go` deletion is safe (it was exported, but only within-repo importable and unused elsewhere).

## Correctness analysis of the focus points

- **Dispatch goroutine lifetime**: one goroutine per *successful* `start()`, guaranteed single by the `hc.me` guard under `hc.lock`; it exits when the subscribe channel closes. Given the client is a process-lifetime singleton and no `Shutdown` exists today (the commented-out one predates this PR), this is the right scope — no per-call growth, nothing to leak repeatedly. When #362 gives the client a real lifecycle, `Close` should close the subscription so dispatch exits; fine to defer.
- **Pending-map double-delete**: safe by construction. `dispatch` removes the entry under `pendingLock` *before* sending, so at most one send ever targets a given `respCh`; `respCh` is buffered (1), so a send racing a timed-out caller never blocks dispatch; `CallE`'s deferred `delete` on an already-removed key is a no-op. A duplicate/late response finds no entry and is logged and dropped. No deadlock, no leak, no double-send.
- **Registration-before-publish** is correct and the comment explains why — dispatch cannot race ahead of registration.
- **`start()` state commit**: all fields (`me`, `from`, `to`, `pending`) assigned only after both Subscribe and Publisher succeed, so a failed attempt is fully retryable; every `CallE` passes through `hc.lock` in `start()`, which also gives the necessary happens-before for `hc.to`/`hc.from` reads. Matches the issue's prescribed fix exactly.
- **Receiver-vs-global**: no `TheClient` reference remains in any method of `client` (verified by grep); the redundant `start()` pre-calls in `LookupDevices`/`ForgetDevices` were correctly dropped (CallE starts lazily). Acceptance criterion met.
- **Tests** hit all three acceptance criteria precisely: reversed-order concurrent responses, prompt error on Subscribe failure, poisoned `TheClient` with `t.Cleanup` restore and no `t.Parallel()` (per CLAUDE.md's RPC-handler test rule). The fake transport is a nice deterministic design.

## Timeout-semantics change — one recommended tweak (non-blocking)

`hc.timeout` now starts *after* the publish handoff; the publish send is guarded only by `ctx`. That is strictly better than the old unguarded `hc.to <- reqStr` (the issue's own prescribed fix, implemented as written), but it leaves one unbounded path: with a no-deadline ctx (daemon-internal calls) and a full/undrained publisher channel (broker down), `CallE` blocks until ctx cancellation rather than failing after `hc.timeout`. Two-line improvement: create `timeoutCtx` *before* the publish select and use `timeoutCtx.Done()` in both selects, so `hc.timeout` bounds the whole call. Fine to do here or fold into #364 (robustness/lifecycle), whose scope it naturally matches — flagging for the orchestrator either way.

## Minor notes (no action needed)

- Partial-failure corner: if Subscribe succeeds but Publisher fails, the just-created subscription dangles (the interface has no Unsubscribe today); a retry then calls Subscribe for the same topic again. Rare (both ride the same connection) and not fixable cleanly until the MQTT-interface work (#363) — noting for the record.
- Test dispatch goroutines live until process exit (respCh never closed) — harmless in tests, resolved for real code by the future `Close`.
- The commented-out `Shutdown()` block above `dispatch` is pre-existing §8.6 material — #358's sweep should take it.

All three live bugs fixed as specified, acceptance criteria demonstrably met, race-clean at count=20, lint-clean. Approving.

PR #375, opened 2026-07-14
ci: golangci-lint + go vet + -race jobs (#355)

asnowfix:
Closes #355

## Approach

**1. `.golangci.yml` (v2 schema) at repo root.** Curated linter set per the issue: `govet`, `staticcheck`, `errcheck`, `errorlint`, `unused`, `ineffassign`, `revive`, plus `gofmt`/`goimports` as formatters (in golangci-lint v2, `gci`/`goimports` are formatters, not linters — see Deviations). Pragmatic exclusions, each with an in-file justification: the `std-error-handling` preset (conventional ignored `Close`/`Flush`/`Fprint*` errors), errcheck off in `_test.go` files, and errcheck exclusions for goja emulator `Object.Set`/`Runtime.Set` bindings.

**2. `make lint` and `make test-race` targets.** Both reuse the same `go.work` module loop as `make test`, so all ~59 workspace modules are covered (golangci-lint has no native workspace mode; a naive run covers only the root module). `lint` skips modules with no Go files (`pkg/shelly/blu` is an empty module — go.mod/go.sum only).

**3. CI jobs in `.github/workflows/test.yml`.** New `lint` and `race` jobs run `make lint` / `make test-race`, so local and CI behavior match. Both are independent of the coverage `build` job so its timing is unaffected. golangci-lint is pinned to v2.5.0 via a workflow env var. PRs targeting this integration branch now also trigger the workflow (previously only `main`/`v*.x`), so the roadmap PRs get the same gates.

**4. Baseline cleanup — all findings fixed, none baselined away silently.** ~300 findings across the workspace: mechanical gofmt/goimports + errorlint `%v`→`%w` fixes (~90 files, via `golangci-lint fmt` / `run --fix`); hand-fixed errcheck findings (MQTT publishes now logged or propagated in `internal/myhome/server.go`, `myhome/mqtt/client.go`, `pkg/shelly/gen1/proxy.go`, `myhome/occupancy`; `pprof.StartCPUProfile` checked; CLI `PrintResult`/`os.WriteFile` propagated; deliberate ignores marked `_ =` with comments); staticcheck (deduped double imports of `myhome/mqtt`, un-capitalized error strings, empty branches); unused dead code removed (`statusWriter`, `shellyEventHandler`, `parseKVSValue`, `isTimeoutError`, `splitTimeRange`, `nameRe`, `getEnvironmentStatus`); ineffassign cleanups.

## Verification

- `make generate` / `make build`: OK
[...]

Issue #373, opened 2026-07-14
bug: data race on mockPumpController.calls in solar automation tests (myhome/daemon fails under -race)


asnowfix:
## Context

Found by the new \`go test -race\` CI job introduced for #355 (CI hardening). Reproduce with:

```
cd myhome/daemon && go test -race ./...
```

All 12 \`TestSolarAutomation_*\` tests fail with \`WARNING: DATA RACE\` (plus one hit in \`solar_notice_test.go\`).

## Race

The test fake \`mockPumpController\` records calls in a plain slice with no synchronization (\`myhome/daemon/solar_automation_test.go:15-29\`):

```go
type mockPumpController struct {
	calls []bool // true=on, false=off
}

func (m *mockPumpController) SetPump(_ context.Context, on bool) error {
	m.calls = append(m.calls, on) // written from the automation goroutine
	return nil
}
```

The solar-automation loop invokes \`SetPump\` on its own goroutine while the test goroutine reads \`m.calls\` / \`lastCall()\` (races reported at lines 20, 25, 28, 58, 111, 169, 197, 241, 274, 436, 467, 498; also \`solar_notice_test.go:113\`).

## Suggested fix

Add a \`sync.Mutex\` to \`mockPumpController\` guarding \`calls\` (lock in \`SetPump\`, \`lastCall\`, and any direct \`len(m.calls)\` assertions). Test-only change; production code is not implicated by these reports.

## CI note

Until fixed, the module is listed in the \`RACE_SKIP\` list of the root Makefile's \`test-race\` target (see #355). Remove it from that list when closing this issue.

Related: #355 (CI hardening), umbrella #368.

asnowfix:
## 2026-08-11 — verified fixed on `main`, closing

Verified empirically against `origin/main` at commit `84133f8`, in a clean throwaway worktree
(`git worktree add --detach`, then `make generate`), by running the race detector on the affected
package rather than by reading the diff:

```
cd myhome \&\& go test -race -count=1 ./daemon/...
```

Result: **ok  github.com/asnowfix/home-automation/myhome/daemon  3.496s (no data race reported)**

Grep alone was not treated as sufficient evidence here — an earlier pass in this repo produced two
false readings by mixing working-tree greps with `origin/main` greps while sitting on a different
branch, so the detector output above is the basis for closing.

Closing as fixed. If this resurfaces, reopen with the `-race` output attached; note that
[...]

Issue #365, opened 2026-07-14
errors: standardize %w wrapping, add sentinel errors, remove library panics, log once


asnowfix:
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). Error-handling conventions are inconsistent across the codebase:

1. **Wrapping:** ~263 `fmt.Errorf` calls use `%w`, but a large share still interpolate errors with `%v` or `(%v)` (e.g. `pkg/shelly/device.go:147` `"unable to init MQTT (%v)"`), which severs `errors.Is/As` chains for callers.
2. **No sentinel/typed errors:** callers can't branch on failure kinds. The handful that matter deserve definitions in their owning packages, e.g. `ErrDeviceNotFound`, `ErrNotConnected`, `ErrTimeout`.
3. **Library panics:** `NewServerE` panics when the context lacks a logger (`internal/myhome/server.go:24,48`); `hlog.InitWithLevel` panics if the log writer can't be created (`hlog/hlog.go:67`); `pkg/shelly/mqtt.SetClient` panics on double-set (`pkg/shelly/mqtt/client.go:43`). Libraries should return errors (or fall back, e.g. `logr.Discard()` / stderr); reserve panics for genuine programmer-error invariants.
4. **Log-and-return double reporting:** constructors and RPC paths log an error and then return it (e.g. `myhome/storage/db.go:25-27`, `internal/myhome/client.go` throughout), so every failure appears twice (or more) in logs. Convention: return it; the top level logs once.

## Proposed change

- Enable `errorlint` in golangci-lint (see CI hardening issue) and fix all `%v`-on-error findings to `%w`.
- Introduce the small set of sentinel errors where callers actually branch today (search for string-matching on error text to find them).
- Replace the three library panics with error returns/fallbacks.
- One-page convention note in AGENTS.md: wrap with `%w` + context; log at the outermost caller only; no panics in library code.

## Acceptance criteria

- [ ] `errorlint` passes repo-wide with no suppressions for `%v`-wrapping.
- [ ] The listed panics are gone; affected constructors return errors.
- [ ] At least the device-lookup path exposes a sentinel (`errors.Is(err, ErrDeviceNotFound)`) with a test.

## Dependencies

- Depends on: CI hardening (errorlint enforcement is the ratchet that keeps this fixed).

**Issue links:** depends on #355 (errorlint is the enforcement ratchet).

Issue #355, opened 2026-07-14
ci: add golangci-lint, go vet, and race-detector jobs (CI hardening)


asnowfix:
## Context

From a full Go expert review of the codebase (2026-07-13, main @ 2cd3628). The repository currently has **no static analysis at all**: no `.golangci.yml`, no `go vet` step, and no race-detector run in CI. `.github/workflows/test.yml` runs `make cover` (tests + coverage gate) only.

This is the highest-leverage cheap fix identified by the review. Defects that a standard linter setup finds mechanically are present in the code today, e.g.:

- A verbatim-duplicated CPU-profile block in `myhome/main.go:52-62` and `70-80` (dead/duplicate code).
- Ignored error returns: `pprof.StartCPUProfile(f)` (myhome/main.go), `mc.Publish(...)` in `internal/myhome/server.go:124,145`, `json.Marshal` in `server.fail()`.
- Dead assignment `res.Result = &tempResult` immediately overwritten (`internal/myhome/server.go:93,114`).
- `fmt.Errorf` wrapping with `%v` instead of `%w`, breaking `errors.Is/As` (e.g. `pkg/shelly/device.go:147`).
- Unsorted/ungrouped imports (stdlib mixed with third-party, e.g. `internal/myhome/client.go`).

The codebase also has ~36 raw `go func` sites and known cross-goroutine shared state; CI never runs `go test -race`.

## Proposed change

1. Add `.golangci.yml` at the repo root enabling at least: `govet`, `staticcheck`, `errcheck`, `errorlint`, `unused`, `ineffassign`, `gci` (or `goimports`), `revive`.
2. Add a lint job to `.github/workflows/test.yml` (or a new workflow) running `golangci-lint` across **all workspace modules** (the repo is a multi-module workspace; a naive run covers only the root module — loop over modules like `make test` does).
3. Add a `-race` test job: same module loop, `go test -race ./...` per module. Keep it as a separate job so the coverage job's timing is unaffected.
4. Fix (or explicitly `//nolint` with justification) the findings the linters raise. It is acceptable to start with a curated linter set and a clean baseline rather than enabling everything at once.
5. Wire the same entry points into the `Makefile` (`make lint`, `make test-race`) so local and CI behavior match.

## Why first

This issue is the foundation for the rest of the review's refactoring plan (RPC-core fixes, module collapse, dependency-injection pass): landing linters *before* those refactors means every subsequent PR is checked against the hardened baseline.

## Acceptance criteria

[...]

PR #380, opened 2026-07-17
fix(main): profile file lifecycle, cancel-func plumbing, daemon detection (#357)

### LINKED ISSUE #357: cli: duplicated CPU-profile block in main.go; profiling & shutdown error handling
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). `myhome/main.go` has a copy-paste bug and several error-handling/context-usage problems around CPU profiling and shutdown.

## Problems

1. **Duplicated CPU-profile block.** `PersistentPreRunE` contains the same block twice (`myhome/main.go:52-62` and `70-80`), and each copy contains a nested duplicate `if options.Flags.CpuProfile != ""` inside itself. With `--cpuprofile` set: `os.Create` runs twice (first file handle leaked), `pprof.StartCPUProfile` runs twice (second call errors — ignored), and the context value is overwritten.
2. **`pprof.StartCPUProfile(f)` error is never checked.**
3. **The open `*os.File` is stored in `context.WithValue` and never closed.** `PersistentPostRunE` (`myhome/main.go:96-99`) checks `f != nil` only to execute `defer pprof.StopCPUProfile()` — a `defer` as the last statement of a function is just a call, and the file is still not closed/flushed, so the profile tail can be lost.
4. **Unchecked type assertion on shutdown:** `ctx.Value(global.CancelKey).(context.CancelFunc)` (`myhome/main.go:101`) panics for any code path that didn't stash a cancel func in the context. Same pattern exists in `myhome/daemon/daemon.go:69` (`NewDaemon`).
5. **Daemon detection by name string-matching:** `cmd.Name() == "daemon" || cmd.Parent().Name() == "daemon"` (`myhome/main.go:29`) breaks on deeper command nesting. Prefer `cmd.Annotations` or a flag set by the daemon command itself.

## Proposed fix

- Keep one profile block; check the `StartCPUProfile` error; hold the file in a package-level var (or closure) in `main.go`; in post-run: `pprof.StopCPUProfile()` then `f.Close()`, both checked.
- Replace the context-value cancel-func plumbing with an explicitly owned `context.WithCancel` in `main`/daemon (context values are for request-scoped data, not for control-flow dependencies).
- Replace name sniffing with an annotation on the daemon command.

## Acceptance criteria

- [ ] `myhome --cpuprofile /tmp/p.out ctl ...` produces a valid, complete profile (verify with `go tool pprof`).
- [ ] No `context.WithValue` carries `*os.File` or `context.CancelFunc`.
- [ ] `myhome daemon run` and nested daemon subcommands still get daemon logging defaults (verbose-by-default), covered by a test on the annotation logic.

asnowfix:
Closes #357

## Approach

Implements the remaining scope of #357 (see deviation note below):

- **CPU-profile file lifecycle** (`myhome/main.go`): the profile `*os.File` no longer travels through `context.WithValue`; it lives in a package-level var set once in `PersistentPreRunE` and consumed once in `PersistentPostRunE`, which now calls `pprof.StopCPUProfile()` then `f.Close()` with the close error checked and logged. Previously the last statement was a no-op trailing `defer` and the file was never closed, so the profile tail could be lost. On `StartCPUProfile` failure the file is closed instead of leaked.
- **Cancel-func plumbing** (`myhome/ctl/options/options.go`, `myhome/main.go`, `myhome/ctl/ctl.go`, `myhome/daemon/daemon.go`): `options.CommandLineContext` now returns `(context.Context, context.CancelFunc)` instead of stashing the CancelFunc in the context. Each entry point owns its CancelFunc explicitly, and `daemon.NewDaemon` derives its own `context.WithCancel(ctx)` — parent cancellation still propagates, so SIGINT/SIGTERM handled upstream and service `Stop()` both converge on the same `<-d.ctx.Done()` wait in `Run`. This removes both panic-prone unchecked `ctx.Value(global.CancelKey).(context.CancelFunc)` assertions (`myhome/main.go`, `myhome/daemon/daemon.go`). `global.CancelKey` and `global.CpuProfileKey` are deleted: no `context.WithValue` carries an `*os.File` or `context.CancelFunc` anymore.
- **Daemon detection** (`myhome/daemon/cmd.go`): name string-matching (`cmd.Name() == "daemon" || cmd.Parent().Name() == "daemon"`, which breaks on deeper nesting) replaced with `IsDaemonAnnotation` set on `daemon.Cmd` plus `daemon.IsDaemonCommand(cmd)`, which walks the ancestor chain. Covered by a new table test in `myhome/daemon/cmd_test.go` (3-level nesting, sibling trees, nil, and the real run/install/uninstall commands).
- `cmd/datacollector/main.go` adapted to the new `CommandLineContext` signature (`defer cancel()`).

## Deviation note

Problems 1–2 of the issue (duplicated CPU-profile block, unchecked `pprof.StartCPUProfile` error) were **already fixed** in 25bf682 (PR #375, part of #355's errcheck cleanup). This PR implements only the remaining problems 3–5.

## Verification

- `make generate`, `make build`, `make test` — all pass.
- `golangci-lint run` and `go vet` — 0 issues on all touched packages.
- `go test -race ./myhome/daemon/...` — only the known pre-existing #373 race fails (solar automation mock, untouched here); the new `TestIsDaemonCommand` passes under `-race`; other touched packages race-clean.
- Acceptance criteria: `myhome --cpuprofile <file> version` produces a profile that `go tool pprof -top` parses cleanly (valid header, correct duration); grep confirms no `context.WithValue` stores `*os.File`/`CancelFunc`; annotation logic covered by test. `daemon run --help`, `daemon install --help`, `ctl list --help` all exit 0.

## Out-of-scope findings (documented on #357, not fixed here)

- `myhome/ctl/ctl.go`'s cleanup goroutine calls `os.Exit(0)` on signal, skipping `PersistentPostRunE` — a Ctrl-C'd `ctl` run still truncates its CPU profile; the goroutine also double-handles signals already handled by `options.CommandLineContext`. Belongs with the goroutine-lifecycle work in #364.
- `daemon.Run` starts the pprof HTTP server with timeout-less `http.ListenAndServe` — already covered by #364.
- `cmd/datacollector/main.go` uses `log.Fatalf` throughout, which skips deferred cleanup — #365 material.
- Confirmed the #373 race is still live (not RACE_SKIP'd) on this integration branch.

## Commits

- fix(main): stop leaking CPU-profile file and panicking on shutdown ([d9c5eba](https://github.com/asnowfix/home-automation/commit/d9c5ebaeab91d723f16310beeab6c8f62936f353))

### DIFF
--- a/cmd/datacollector/main.go
+++ b/cmd/datacollector/main.go
@@ -63,7 +63,8 @@ func main() {
 	logger = hlog.Logger
 
 	ctx := logr.NewContext(context.Background(), logger)
-	ctx = options.CommandLineContext(ctx)
+	ctx, cancel := options.CommandLineContext(ctx)
+	defer cancel()
 
 	mc, err := mqtt.GetClientE(ctx)
 	if err != nil {
--- a/internal/global/global.go
+++ b/internal/global/global.go
@@ -7,9 +7,12 @@ import (
 type ContextKey uint
 
 const (
-	CancelKey ContextKey = iota
-	CpuProfileKey
-	VersionKey
+	// VersionKey and ProcessContextKey carry request-scoped data that's
+	// legitimately read far from where it's set. Control-flow dependencies
+	// (a CancelFunc, an *os.File) are NOT stored here — see myhome/main.go,
+	// myhome/ctl/ctl.go and myhome/daemon/daemon.go, which each own their
+	// cancellation/profile lifecycle explicitly instead.
+	VersionKey ContextKey = iota
 	ProcessContextKey
 )
 
--- a/myhome/ctl/ctl.go
+++ b/myhome/ctl/ctl.go
@@ -4,11 +4,9 @@ import (
 	"context"
 	"os"
 	"os/signal"
-	"runtime/pprof"
 	"syscall"
 
 	"github.com/asnowfix/home-automation/hlog"
-	"github.com/asnowfix/home-automation/internal/global"
 	"github.com/asnowfix/home-automation/internal/myhome"
 	"github.com/asnowfix/home-automation/myhome/ctl/blu"
 	"github.com/asnowfix/home-automation/myhome/ctl/config"
@@ -37,6 +35,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// cancelFunc is owned by this package alone, set once in PersistentPreRunE
+// and consumed once in PersistentPostRunE. It's a package-level var rather
+// than a context value on purpose: context.WithValue is for request-scoped
+// data, not for a control-flow handle like a CancelFunc (an unchecked type
+// assertion on a missing/mistyped value would panic). ctl.Cmd runs once per
+// process invocation, so there's no concurrent-invocation hazard here.
+var cancelFunc context.CancelFunc
+
 var Cmd = &cobra.Command{
 	Use:              "ctl",
 	Short:            "Control and manage home automation devices",
@@ -50,7 +56,9 @@ var Cmd = &cobra.Command{
 		log := hlog.Logger
 		ctx := logr.NewContext(cmd.Context(), log)
 
-		ctx = options.CommandLineContext(ctx)
+		var cancel context.CancelFunc
+		ctx, cancel = options.CommandLineContext(ctx)
+		cancelFunc = cancel
 
 		// Set the target instance name for RPC topics
 		if options.Flags.InstanceName != "" {
@@ -109,18 +117,18 @@ var Cmd = &cobra.Command{
 	PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
 		ctx := cmd.Context()
 
-		f := ctx.Value(global.CpuProfileKey)
-		if f != nil {
-			defer pprof.StopCPUProfile()
-		}
+		// CPU profile start/stop/close is owned entirely by myhome/main.go's
+		// PersistentPostRunE (which always runs, as the root command's
+		// persistent hooks apply to every subcommand); nothing to do here.
 
 		mc, err := mqttclient.GetClientE(ctx)
 		if err != nil {
 			return err
 		}
 
-		cancel := ctx.Value(global.CancelKey).(context.CancelFunc)
-		cancel()
+		if cancelFunc != nil {
+			cancelFunc()
+		}
 		mc.Close()
 		return nil
 	},
--- a/myhome/ctl/options/options.go
+++ b/myhome/ctl/options/options.go
@@ -111,7 +111,13 @@ var Flags struct {
 
 var Via types.Channel
 
-func CommandLineContext(ctx context.Context) context.Context {
+// CommandLineContext builds the cancellable, request-scoped context shared by
+// every myhome entry point (daemon and ctl). It returns the derived context
+// together with its CancelFunc: the caller owns that CancelFunc explicitly
+// (e.g. in a package-level var used from PersistentPostRunE) rather than
+// threading it through context.WithValue, which is for request-scoped data,
+// not control-flow dependencies.
+func CommandLineContext(ctx context.Context) (context.Context, context.CancelFunc) {
 	var cancel context.CancelFunc
 
 	// Create the process-wide context that background services can use
@@ -119,10 +125,8 @@ func CommandLineContext(ctx context.Context) context.Context {
 
 	if Flags.Wait > 0 {
 		ctx, cancel = context.WithTimeout(processCtx, Flags.Wait)
-		ctx = context.WithValue(ctx, global.CancelKey, cancel)
 	} else {
 		ctx, cancel = context.WithCancel(processCtx)
-		ctx = context.WithValue(ctx, global.CancelKey, cancel)
 	}
 
 	// Store the process-wide context so lazy-started services can access it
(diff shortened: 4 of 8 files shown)


asnowfix (commented):
## verdict: APPROVE

(gh is authenticated as the PR author, so this is a COMMENT review carrying an explicit approval verdict.)

Reviewed d9c5eba against #357's remaining scope (problems 3–5). Independently verified in a scratch worktree; all claims check out.

### Verified

- **Call-site completeness**: `CommandLineContext`'s new `(ctx, cancel)` signature has exactly three callers (`myhome/main.go`, `myhome/ctl/ctl.go`, `cmd/datacollector/main.go`), all updated. Zero remaining references to `global.CancelKey`/`CpuProfileKey`; both panic-prone unchecked type assertions are gone.
- **Cobra hook chaining**: `cobra.EnableTraverseRunHooks = true` (myhome/main.go:132) confirms root `PersistentPostRunE` runs for every subcommand, so centralizing profile stop/close there (and deleting ctl.go's no-op `defer pprof.StopCPUProfile()`) is correct. The `<-ctx.Done()` in root PostRunE cannot deadlock: when PostRunE runs, PreRunE succeeded, so `shutdownCancel` is always non-nil and called first.
- **Daemon shutdown ordering preserved**: every service the daemon starts derives from `d.ctx`, and `Run` waits on `<-d.ctx.Done()`, so `Stop()` cancelling the new child context reaches everything it did before; parent (signal) cancellation still propagates down. `processCtx` was never cancelled by `Stop()` in either version — no change. `install`/`uninstall` previously depended on root PreRunE having stashed `CancelKey`; the self-owned `WithCancel` removes that hidden coupling with identical behavior. Daemon-optional resilience unaffected.
- **Annotation coverage**: `run`, `install`, `uninstall` are all `AddCommand`'d under `daemon.Cmd`, which carries the annotation; `IsDaemonCommand`'s ancestor walk covers arbitrary nesting and nil. The table test asserts the real commands, not just synthetic trees — good.
- **Local gate** (see CI note below): `make generate && make build` pass; `go vet` and `golangci-lint run` on `./myhome/... ./cmd/... ./internal/global/...` — 0 issues; `go test` on those trees passes; `TestIsDaemonCommand` clean under `-race`.
- **End-to-end acceptance**: `myhome --cpuprofile p.out version` and `myhome --cpuprofile p.out ctl --mqtt-broker tcp://192.168.1.2:1883 list` (live broker, successful run) both produced non-empty profiles that `go tool pprof -top` parses cleanly, samples included.

### Non-blocking notes

1. **Three new package vars, not one** (`cpuProfileFile` + `shutdownCancel` in main, `cancelFunc` in ctl). All are set-once/consumed-once per process with clear ownership comments; appropriate for a CLI that runs `Execute()` once. Fine as is.
2. **ctl.go PostRunE ordering nit**: `mqttclient.GetClientE(ctx)` returning an error exits before `cancelFunc()` runs, leaking the cancel on that error path (process exits anyway, so cosmetic). Calling `cancelFunc()` before the client lookup — or deferring it — would be marginally tighter. Not worth a round-trip.
3. **Error-path profile truncation** (pre-existing, out of scope): a *failed* ctl run (e.g. broker unreachable) leaves a 0-byte profile because cobra skips PostRun hooks when PreRunE/RunE errors — same family as the `os.Exit(0)` signal-path gap already documented on #357 for the #364 goroutine-lifecycle work. Verified empirically.

### Out-of-scope defect found during review (for the orchestrator, needs an issue)

**The Build and Test CI gate has never run on integration-branch PRs.** `test.yml`'s `pull_request.branches` entry `worktree-feature+golang-expert-review` (added in PR #375 for exactly this purpose) contains `+`, which is a *special character* in GitHub Actions filter patterns (one-or-more quantifier), so the pattern never matches the literal branch name. Verified: no "Build and Test" workflow run exists for the head commits of PR #378 or #380 — only advise/merge-policy/description jobs ran. Fix: escape it (`worktree-feature\+golang-expert-review`, single-quoted in YAML). Until fixed, none of the campaign PRs are CI-gated; this PR was gated locally instead (see above).

asnowfix:
Follow-up on the CI-gate finding in my review, now confirmed empirically across every integration-branch PR:

| PR | base | Build and Test run? |
|---|---|---|
| #375 | worktree-feature+golang-expert-review | none (only advise/merge-policy/update-pr-description/auto-merge) |
| #378 | worktree-feature+golang-expert-review | none |
| #380 | worktree-feature+golang-expert-review | none |

Meanwhile "Build and Test" runs normally for `pull_request` events based on `main` and `v0.11.x` (e.g. run 29563373940 for worktree-bugfix+ui-duplicated-devices, triggered the same minute #380 was opened). Note that #375 itself carried the updated workflow file — `pull_request` events use the workflow from the PR's merge ref — so the filter had every chance to match its own PR and did not. Conclusion: the unescaped `+` in the `pull_request.branches` entry of `.github/workflows/test.yml` never matches the literal branch name (`+` is a one-or-more quantifier in Actions filter patterns). Suggested fix, to be validated with a throwaway PR against the integration branch: `- 'worktree-feature\+golang-expert-review'`.

This does not block #380 (gated locally: build, vet, golangci-lint 0 issues, tests, targeted -race all pass), but every past and future campaign PR is un-gated until it's fixed. Filing this as a follow-up issue is the orchestrator's call.

PR #389, opened 2026-07-18
chore: remove committed debug artifacts and dead code (#358)

asnowfix:
Closes #358

## Summary

Removes committed debug artifacts (some containing real LAN network data) and dead commented-out code, per the #358 hygiene findings from the #368 Go expert review.

## Removed files (git rm — none referenced by build/CI/tests/embeds, verified with repo-wide grep)

- `goroutine-dump.txt` — 57 KB pprof goroutine dump
- `data.js`, `data-shelly.js`, `data-list.json` — **contain private LAN data**: real Shelly device IDs (e.g. `shelly1minig3-54320440d02c`) and real LAN IPs (`192.168.1.x`, `192.168.33.x`, `192.168.34.x`)
- `myhome.yaml` — root-level live instance config

This content remains reachable in git history; history rewriting is a separate decision and out of scope here (per the issue text).

## PLAN.md

Not stale — it's the active pkg/shelly → go-shellies extraction plan (feeds #359). Moved with `git mv` (history preserved) to `docs/EXTRACT-PKG-SHELLY-PLAN.md`, matching the project's plan-file convention and the naming of sibling `docs/*-PLAN.md` files.

## .gitignore additions

```
/goroutine-dump*.txt
/data*.js
/data-*.json
/myhome.yaml
```

## Dead code removed

- `pkg/shelly/device.go` — ~30-line commented-out wifi/ethernet/system-config refresh block in `refresh()`
- `internal/myhome/server.go` — commented-out `mc`/`to` fields, publisher wiring, abandoned `reflect.TypeOf` type-check block
- `internal/myhome/client.go` — commented-out `Shutdown()` method (explicitly flagged as in-scope by the #356 review)
- `pkg/devices/types.go` — commented-out `Devices` struct

## Verification

- `make generate` — OK
- `make build` — OK
- `make test` — all packages pass (root + go.work submodules)
- `golangci-lint run` on touched packages (`pkg/shelly/...`, `internal/myhome/...`, `pkg/devices/...`) — 0 issues

## Out-of-scope findings (not fixed here, flagging for a future issue)

In `pkg/shelly/device.go`, adjacent to but not listed in #358:
- `IsHttpReady()` — an unreachable ~17-line commented block after an early `return`, marked `FIXME: check that the IP is on the same network as the Router`
- A fully commented-out `refreshScripts()` method (~30 lines), already orphaned before this cleanup (its only caller was itself commented out inside the block removed here)

## Test plan

- [x] `make generate && make build`
- [x] `make test`
[...]

Issue #372, opened 2026-07-14
bug: data races in script-emulator DeviceState maps (internal/shelly/scripts tests fail under -race)


asnowfix:
## Context

Found by the new \`go test -race\` CI job introduced for #355 (CI hardening). Reproduce with:

```
cd internal/shelly/scripts && go test -race ./...
```

8 tests fail with \`WARNING: DATA RACE\`:

- TestBluListener_MotionTurnsOnSwitch
- TestBluListener_AutoOffTurnsOffSwitch
- TestBluListener_IlluminanceWithinBoundsTurnsOn
- TestBluListener_KVSReloadPicksUpNewFollow
- TestPoolPump_InitVerifiesSchedules
- TestPoolPump_WaterSupplyRestoresSpeed
- TestPoolPump_ButtonCyclesPro3
- TestPoolPump_Pro1ToggleAndWaterSupply

## Race

The script emulator's shared \`DeviceState\` maps are mutated from the goja VM goroutine while the test goroutine reads them without synchronization:

- Write side: \`pkg/shelly/script/run.go:896\` (\`kvs.set\` handler writes \`deviceState.GetKVS()\` map) and \`run.go:992\` (\`switch.set\` handler writes \`deviceState.ComponentStatus\`), reached from timer/event-driven script callbacks running on a different goroutine.
- Read side: test helpers such as \`internal/shelly/scripts/blu_listener_test.go:87\` (\`switchIsOn\` reading \`ComponentStatus\`) and the pool-pump test assertions (\`pool_pump_test.go:444-471\`).

## Suggested fix

Either guard \`DeviceState\` (KVS map, \`ComponentStatus\`) with a mutex inside the emulator (\`pkg/shelly/script\`), or expose snapshot accessors that the tests use instead of reaching into the maps. The emulator is daemon-side (device-extension mode), so a mutex is acceptable — no Shelly-side resource limits apply.

## CI note

Until fixed, the module is listed in the \`RACE_SKIP\` list of the root Makefile's \`test-race\` target (see #355). Remove it from that list when closing this issue.

Related: #355 (CI hardening), #364 (goroutine lifecycle), umbrella #368.

asnowfix:
## 2026-08-11 — verified fixed on `main`, closing

Verified empirically against `origin/main` at commit `84133f8`, in a clean throwaway worktree
(`git worktree add --detach`, then `make generate`), by running the race detector on the affected
package rather than by reading the diff:

```
[...]

Issue #359, opened 2026-07-14
build: collapse 59 Go modules into 4 (root + pkg/shelly, pkg/sfr, pkg/beem)


asnowfix:
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). The repository is split into **59 Go modules** (59 `go.mod` files, ~59 `use` entries in `go.work`) for what is essentially one binary plus a few libraries. Costs, all observed in the repo today:

- Cross-package imports are cross-*module* deps: the root `go.mod` requires its own sub-modules at stale pseudo-versions plus 16 `replace` directives.
- `go build/test/vet/generate ./...` don't cover the tree; the Makefile reimplements recursion, and CLAUDE.md documents a 4-place checklist ("go generate sub-module gap") to keep CI green.
- Module boundaries exist at meaningless granularity (e.g. `myhome/ctl/blu/follow` — a single cobra subcommand — is its own module).
- Dependabot opens one PR per module per dependency.

## Decision (owner-approved)

Collapse **59 modules → 4**: the root app module plus three standalone, exportable library modules:

- **`pkg/shelly`** — Shelly Gen2+ RPC client (MQTT + HTTP); genuine external value (existing Go Shelly libraries are HTTP-only). Currently itself split into 17 modules — becomes ONE module.
- **`pkg/sfr`** — SFR box (French ISP router) local API client incl. auth handshake; no widely-available Go equivalent.
- **`pkg/beem`** — Beem Energy API client.

Everything else folds into the root module, including `pkg/devices` (an interface grab-bag + global provider registry with no external value — interfaces belong at the consumer), `pkg/tapo` (71 lines thinly wrapping `github.com/j-iot/tapo-go`; an external user would import tapo-go directly), `pkg/version` (app ldflags glue), `hlog`, all `internal/*`, all `myhome/*`.

## Collapse order (each step independently green)

1. Merge the 16 `pkg/shelly/*` sub-modules into `pkg/shelly` (delete their `go.mod`/`go.sum`, clean `go.work`, one `go mod tidy`).
2. Sever `pkg/shelly`'s app-coupling imports (tracked in the companion issue "pkg/shelly: decouple from app internals"):
   - blank import `_ ".../internal/myhome/net"` (`pkg/shelly/device.go:7`)
   - `internal/shelly/scripts` import in `pkg/shelly/ops.go`
   - `pkg/devices` dependency (constructors return concrete `*shelly.Device`)
[...]

PR #392, opened 2026-07-19
refactor(shelly): decouple pkg/shelly from app internals, clean public API (#360)

### LINKED ISSUE #360: pkg/shelly: decouple from app internals and clean up public API for standalone export
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). `pkg/shelly` is the repo's flagship library (Shelly Gen2+ RPC over MQTT and HTTP) and is slated to become a standalone exportable module. Its public API has several unidiomatic constructs and app-couplings that should be fixed before (or while) the module boundary becomes real.

## Problems & proposed fixes

1. **App-coupling imports (blockers for standalone export):**
   - `pkg/shelly/device.go:7` blank-imports `internal/myhome/net` for init side effects — a pkg→internal layering violation, invisible to readers. Register whatever it registers explicitly from the app's composition root.
   - `pkg/shelly/ops.go` imports `internal/shelly/scripts` — the catalog of house-specific automation JS (garden, pool-pump, front-door…). A generic Shelly client must not embed one house's scripts: callers pass script sources in (an `Upload(ctx, dev, name, src []byte)`-shaped API already exists).
   - `pkg/shelly` imports `pkg/devices` only for the `devices.Device` interface, a `Resolver` interface, and a JSON helper (`pkg/shelly/device.go:591,628,642`, `mdns.go:18`, `shelly.go:43`). Constructors should return the concrete `*shelly.Device` ("accept interfaces, return structs"); declare the small `Resolver` interface locally at its point of consumption; inline the marshal helper.

2. **Trailing-underscore exported fields.** `Device` exports `Id_`, `MacAddress_`, `Name_`, `Host_` (`pkg/shelly/device.go:117-120`) to dodge collisions with the `Id()`, `Name()`, `Host()` getters. Not a Go pattern from any style guide. Fix: unexported fields + `MarshalJSON`/`UnmarshalJSON` (or a small `deviceJSON` DTO), keeping the methods.

3. **`Device` mixes entity, transport, cache, and logger.** One struct holds identity, live MQTT channels, a `dialogs sync.Map`, cached info/config/status, a `modified` flag, and a logger. This is why the package-level `deviceMqttRegistry` global (`pkg/shelly/device.go:110-114`) exists — Device values recreated from the DB must re-find their transport. Separate the serializable **data** from a **session/transport** object owned by the device manager; the registry becomes an ordinary map inside the manager.

4. **Interface bloat.** `types.Device` has ~25 methods; the `FakeDevice` test double in the same package needs ~30 lines of no-op stubs. Most consumers need only `CallE` + identity accessors. Define narrow interfaces where they are *consumed* and let `*shelly.Device` satisfy them implicitly. Also move `FakeDevice` out of the production package into a `typestest` package (stdlib pattern: `net/http/httptest`).

5. **Hand-maintained parallel enum arrays.** `types.Api` is an iota enum whose `String()` indexes a 23-element array literal (`pkg/shelly/types/types.go:120+`). An insertion in one list without the other silently shifts every name after it. Use `//go:generate stringer -type=Api` or a `map[Api]string` verified by a test.

## Acceptance criteria

- [ ] `go list -deps ./pkg/shelly/...` shows no `internal/*`, `myhome/*`, or `pkg/devices` imports.
- [ ] No exported identifier ends with `_`.
- [ ] `FakeDevice` lives in a test-support package, not in `types`.
- [ ] `Api.String()` cannot silently desynchronize (stringer-generated or map+test).

## Dependencies

- Blocks: the 59→4 module collapse (its step 2 is exactly item 1 above).
- Item 3 (Device split) can land separately; items 1–2 are the export blockers.

**Issue links:** blocks #359 (module collapse, step 2).

asnowfix:
Closes #360. Part of umbrella #368 (Wave 3 — export blocker for the standalone go-shellies extraction, see docs/EXTRACT-PKG-SHELLY-PLAN.md).

## What

- **Zero app-couplings left in `pkg/shelly`**: no `internal/*`, `myhome/*`, or `pkg/devices` imports remain (verified via `go list -deps`). The `internal/myhome/net` blank-import in `device.go` is gone — mDNS resolver registration is now an explicit call at app wiring time.
- **Public API cleanup**: trailing-underscore exported fields (`Id_`, `MacAddress_`, `Name_`, `Host_`) replaced with unexported fields + `MarshalJSON`/`UnmarshalJSON` (wire format unchanged); device constructors return `*shelly.Device` (no premature interface); `Do`/`Foreach` typed on a small local `shelly.Summary` interface.
- **`FakeDevice`** moved (`git mv`) from `pkg/shelly/types` to a dedicated `pkg/shelly/typestest` module (go.work-registered) so the production package carries no test fake.
- **`types.Api.String()`** switched from a positional array literal to a map keyed by the constants, with a guard test preventing silent desync.
- ~40 downstream files (`myhome/ctl/*`, `internal/myhome`, `myhome/daemon`, `myhome/devices/impl`) mechanically updated.

## Verification (by the implementation agent, pre-commit)

- `make generate` / `make build`: green
- `make test`: full workspace loop green (exit 0, every module ok)
- `golangci-lint`: 0 issues on every touched tree
- `go test -race ./pkg/shelly/...`: pass

## Notes

- Implemented by the issue agent (verification completed and reported green); committed/pushed by the coordinator after the agent's connection dropped at the commit step, so the work is a single squashed-style commit rather than the agent's planned logical grouping — immaterial under the campaign's squash-merge policy.
- Wire-format compatibility of the MarshalJSON/UnmarshalJSON replacements is a key review point.

## Commits

- refactor(shelly): decouple pkg/shelly from app internals, clean public API ([32b0ad9](https://github.com/asnowfix/home-automation/commit/32b0ad95fe23e642472870ade6629a3bf75aa086))
- fix(script): make timerHandler.stopped atomic to fix pre-existing race ([5c3f87c](https://github.com/asnowfix/home-automation/commit/5c3f87c7af7ae1843ce9f9feb92a89fbcff4bdbe))

### DIFF
--- a/cmd/datacollector/main.go
+++ b/cmd/datacollector/main.go
@@ -100,15 +100,9 @@ func main() {
 		logger.Info("Testing device", "id", device.Id(), "name", device.Name())
 
 		// Create Shelly device instance
-		shellyDevice, err := shelly.NewDeviceFromSummary(ctx, logger, device)
+		sd, err := shelly.NewDeviceFromSummary(ctx, logger, device)
 		if err != nil {
 			logger.Error(err, "Failed to create Shelly device", "device_id", device.Id())
-			continue
-		}
-
-		sd, ok := shellyDevice.(*shelly.Device)
-		if !ok {
-			logger.Error(fmt.Errorf("not a Shelly device"), "Invalid device type", "device_id", device.Id())
 			continue
 		}
 
--- a/go.work
+++ b/go.work
@@ -58,6 +58,7 @@ use (
 	./pkg/shelly/sswitch
 	./pkg/shelly/system
 	./pkg/shelly/types
+	./pkg/shelly/typestest
 	./pkg/shelly/wifi
 	./pkg/tapo
 	./pkg/version
--- a/internal/myhome/device.go
+++ b/internal/myhome/device.go
@@ -155,14 +155,10 @@ func (d *Device) Refresh(ctx context.Context) (bool, error) {
 		sd, ok := d.Impl().(*shellyapi.Device)
 		if !ok || sd == nil {
 			var err error
-			device, err := shellyapi.NewDeviceFromSummary(ctx, d.log, d)
+			sd, err = shellyapi.NewDeviceFromSummary(ctx, d.log, d)
 			if err != nil {
 				d.log.Error(err, "Failed to create device from summary", "device", d.DeviceSummary)
 				return false, err
-			}
-			sd, ok = device.(*shellyapi.Device)
-			if !ok {
-				return false, fmt.Errorf("device is not a Shelly device: %T", device)
 			}
 			d.WithImpl(sd)
 		}
--- a/internal/myhome/foreach.go
+++ b/internal/myhome/foreach.go
@@ -3,7 +3,6 @@ package myhome
 import (
 	"context"
 
-	"github.com/asnowfix/home-automation/pkg/devices"
 	"github.com/asnowfix/home-automation/pkg/shelly"
 	"github.com/asnowfix/home-automation/pkg/shelly/types"
 
@@ -15,14 +14,22 @@ import (
 // The function should return a result and an error.
 // The Foreach function aggregates the results of all the function calls and returns a single result.
 // If any of the function calls return an error, the Foreach function will return that error.
-func Foreach(ctx context.Context, log logr.Logger, name string, via types.Channel, fn func(ctx context.Context, log logr.Logger, via types.Channel, device devices.Device, args []string) (any, error), args []string) (any, error) {
+func Foreach(ctx context.Context, log logr.Logger, name string, via types.Channel, fn shelly.Do, args []string) (any, error) {
 	// Get a list of devices that match the given name.
-	devices, err := TheClient.LookupDevices(ctx, name)
+	found, err := TheClient.LookupDevices(ctx, name)
 	if err != nil {
 		return nil, err
 	}
 
+	// pkg/devices.Device already satisfies shelly.Summary structurally, but
+	// Go does not implicitly convert []devices.Device to []shelly.Summary —
+	// each element must be reboxed into the interface pkg/shelly expects.
+	summaries := make([]shelly.Summary, len(*found))
+	for i, d := range *found {
+		summaries[i] = d
+	}
+
 	// Call the given function for each device in the list.
-	out, err := shelly.Foreach(ctx, log, *devices, via, fn, args)
+	out, err := shelly.Foreach(ctx, log, summaries, via, fn, args)
 	return out, err
 }
--- a/myhome/ctl/blu/follow/blu.go
+++ b/myhome/ctl/blu/follow/blu.go
@@ -13,7 +13,6 @@ import (
 	mhblu "github.com/asnowfix/home-automation/internal/myhome/blu"
 	mhscript "github.com/asnowfix/home-automation/internal/myhome/shelly/script"
 	"github.com/asnowfix/home-automation/myhome/ctl/options"
-	"github.com/asnowfix/home-automation/pkg/devices"
 	"github.com/asnowfix/home-automation/pkg/shelly"
 	"github.com/asnowfix/home-automation/pkg/shelly/kvs"
 	"github.com/asnowfix/home-automation/pkg/shelly/types"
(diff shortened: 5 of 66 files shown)


asnowfix (commented):
## verdict: APPROVE — with one merge condition (pre-existing race, see below)

(gh is authenticated as the PR author, so this is a COMMENT review carrying an explicit approval verdict. It also serves as the required manual review for the `merge-policy` size gate.)

Reviewed 32b0ad9 in a scratch worktree, re-verifying the final tree state independently (the implementer's claims were pre-commit).

### Verified

1. **Wire-format compatibility — empirically proven byte-identical.** On the base commit I marshaled `shelly.Device{Id_, Name_, Host_}` with the old default marshaling: `{"id":"shellypro3-a0dd6ca1c588","name":"pool","host":"192.168.1.55"}`. On the PR tree, the new `UnmarshalJSON` accepts exactly those bytes, hydrates `Id()/Name()/Host()` correctly, and `MarshalJSON` re-emits them byte-for-byte. `MacAddress_` was already `json:"-"` before, so its exclusion from `deviceJSON` is also compatible. The classic hazard of introducing `MarshalJSON` — method promotion through embedding swallowing an outer struct's fields — is absent: no struct in the repo embeds `shelly.Device`. The inlined `ShellyDevice.MarshalJSON` helper reproduces the old `devices.MarshalJSON` field set, tags, and order exactly. Stored `myhome.db` rows, MQTT payloads, and the UI are safe.
2. **mDNS blank import: proven vestigial, removal behavior-neutral.** `internal/myhome/net` contains **no `func init`** at any commit — the blank import in `device.go` was residue of the commented-out `mynet.IsSameNetwork` call (whose comment corpse #389 just deleted). There was no side effect to replace. The explicit seams already existed and are intact: `shelly.SetHostResolver` installed in `DeviceManager.Start` (unchanged), and `NewDeviceFromZeroConfEntry` takes an explicit `Resolver` (sole caller `myhome/daemon/watch/zeroconf.go:81` passes one). The scripts decoupling adds `scriptsFS fs.FS` to `shelly.Init`, nil-tolerant and documented; both entry points pass `scripts.GetFS()` — `daemon.Run` and ctl's `PersistentPreRunE`, the latter covering the MCP server path (a ctl subcommand). `cmd/datacollector` never called `Init` before or after.
3. **No pkg→app leaks.** `go list -deps ./pkg/shelly/... ./pkg/shelly/typestest/...` on the final tree: zero `asnowfix/home-automation` `internal/*`, `myhome/*`, or `pkg/devices` packages. Acceptance criterion met.
4. **typestest wiring.** `git mv` rename detected (`pkg/shelly/{types => typestest}/fake_device.go`); registered in `go.work`; the Makefile `test`/`lint`/`cover` loops are `go.work`-driven (awk over `use` entries) so the new module is picked up automatically; imported only from sibling modules' `_test.go` files and resolved via the workspace (CI build/lint/coverage green confirms).
5. **API cleanup.** No exported `_`-suffixed identifiers remain (only test function names). `Do`/`Foreach` typed on the local `shelly.Summary` interface with a good doc comment; constructors return `*shelly.Device` — datacollector's now-redundant type assertion correctly deleted. `Api.String()` map + guard tests present; unknown values return `Api(9999)` without panicking.
6. **Local re-verification of the final tree**: `make generate`/`make build` green; `go test -race` on `pkg/shelly`, `pkg/shelly/types` green; downstream touched trees (`myhome/devices/...`, `internal/myhome/...`, `cmd/...`) test green. CI on the final commit: build, lint, coverage-delta, build-pr-deb all pass.

### Merge condition: the red `race` job — pre-existing, NOT this PR

`TestTimerHandlerRecurring` in `pkg/shelly/script` fails `-race`: unsynchronized `timerHandler.stopped` bool, written by `Stop()` (`run.go:1245`) and read by the `Wait` goroutine (`run.go:1197`). Two proofs it's not this PR: (a) 32b0ad9 touches **nothing** under `pkg/shelly/script/` (`git show --stat` is empty for that path); (b) the race reproduces on the base commit df9e032 with `go test -race -run TestTimerHandler -count=30`. It is distinct from the known #372/#373/#374. Per campaign protocol this is an out-of-scope defect for the orchestrator: file an issue + `RACE_SKIP` it (or just fix it — `stopped` as `sync/atomic.Bool` is a ~3-line change), then re-run the race job. **Do not merge on a red race job**, even an unrelated one — the gate's value is that red always means stop.

### Non-blocking notes

- **Residual export blocker for the extraction plan**: `pkg/shelly/script/ops.go:42` still imports the app's `hlog` module (`hlog.GetLogger`). It passes #360's acceptance letter (not `internal/*`/`myhome/*`/`pkg/devices`), but #359 folds `hlog` into the root module, at which point the planned CI boundary guard must reject it. Track it in docs/EXTRACT-PKG-SHELLY-PLAN.md (currently absent there) — same treatment as `pkg/beem`'s logger-as-parameter step.
- **Api guard has one hole**: a constant added to the iota block but omitted from *both* `apiNames` and the test's hand-copied `all` slice passes undetected (both length sides move in lockstep, i.e. not at all). A trailing sentinel constant (`apiCount Api = iota` at block end) plus `len(apiNames) == int(apiCount)` closes it and removes the hand-maintained list.
- Two stale comments still reference `Host_` (`pkg/shelly/device.go:399,449`) — sweep opportunistically.
- Sibling `pkg/shelly/*` module `go.mod`s carry no `require` for typestest (workspace resolves it). Irrelevant once #359 collapses them; noting for completeness.

PR #400, opened 2026-07-27
rpc: generics-based method registry + json.RawMessage wire types (#361)

### LINKED ISSUE #361: rpc: generics-based method registry + json.RawMessage wire types (replace signatures map & double unmarshal)
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). The myhome RPC method registry (`internal/myhome/methods.go`) is a ~270-line map of `func() any` constructors (`signatures`), with runtime `reflect.TypeOf` checks in the client (`internal/myhome/client.go:154`) and a documented 4-step ritual in CLAUDE.md for adding a method (Verb constant → request/response types → signatures map → RegisterMethodHandler). The server double-unmarshals every request and the client marshal/round-trips every result to coerce types.

### 1. Generic registration replaces the signatures map

```go
func Register[P, R any](verb Verb, h func(ctx context.Context, p P) (R, error)) {
    methods[verb] = &Method{
        Name: verb,
        Action: func(ctx context.Context, raw json.RawMessage) (any, error) {
            var p P
            if err := json.Unmarshal(raw, &p); err != nil { return nil, err }
            return h(ctx, p)
        },
    }
}
```

Registration becomes one line per method; parameter types are compile-time checked; the `signatures` map and the `reflect.TypeOf` runtime check disappear; the 4-step checklist becomes 2 steps (verb + `Register`). A typed caller-side helper gives the same benefit to clients:

```go
func Call[P, R any](ctx context.Context, c Client, verb Verb, p P) (R, error)
```

### 2. `json.RawMessage` wire types replace double (un)marshalling

Today: the server unmarshals the full message once to read the method, then unmarshals the **whole message again** into typed params (`internal/myhome/server.go:62,85`); the client re-marshals `res.Result` and re-unmarshals it into the typed result (`internal/myhome/client.go:206-213`). Declaring `Params json.RawMessage` / `Result json.RawMessage` in the `request`/`response` structs lets each side unmarshal each payload exactly once, directly into the typed value.

### 3. Server loop hardening (same files, same PR welcome)

- `NewServerE` panics when the context has no logger (`internal/myhome/server.go:24,48`) — fall back to `logr.Discard()` or take the logger as a parameter.
- The server processes messages strictly sequentially in one goroutine — a slow handler head-of-line-blocks every other RPC including the UI. Dispatch each message to a (bounded) goroutine.
- `mc.Publish` errors are ignored in the response and `fail` paths (`server.go:124,145`); `fail` also ignores its `json.Marshal` error.
- Delete the dead `res.Result = &tempResult` assignment (`server.go:93`) and the commented-out reflect block (`server.go:108-112`).

## Migration note

RPC wire format is unchanged (same JSON, same topics) — only the in-process representation moves to `json.RawMessage`. CLI and daemon must still be release-compatible across versions per the project's rolling-upgrade practice.

## Acceptance criteria

- [ ] `signatures` map deleted; every verb registered via the generic `Register`.
- [ ] Exactly one `json.Unmarshal` per request on the server; no marshal→unmarshal round-trip of results on the client.
- [ ] A test proves a slow handler does not block an unrelated concurrent RPC.
- [ ] CLAUDE.md "Adding a method requires four steps" section updated.

## Dependencies

- Depends on: the RPC client correctness fixes (response correlation) — build the dispatcher first, then retype it; doing both at once risks a big-bang PR.

**Issue links:** depends on #356 (build the response dispatcher first).

asnowfix:
Closes #361. Part of umbrella #368 (Wave 3).

## What

- **Core registry rewrite** (`internal/myhome/{methods,proxy,server,client}.go`): `Register[P, R any]` replaces the `signatures` map of `func() any` constructors; `request.Params`/`response.Result` become `json.RawMessage`, eliminating the double-unmarshal/marshal round-trip. `server.go` now does a single-unmarshal decode with bounded per-message goroutine dispatch (`maxConcurrentRPCs`) instead of blocking the dispatch loop per request; `client.go` drops the `reflect.TypeOf` runtime check. The #356 concurrency work (Dialog.Id correlation, dispatch goroutine, ctx-guarded timeout) is preserved and re-verified under `-race`, plus a new `TestServer_SlowHandlerDoesNotBlockConcurrentRPC`.
- **Mechanical migration**: ~30 `RegisterMethodHandler` call sites → `Register`, ~40 `TheClient.CallE` call sites → the generic `Call[P, R]` wrapper, across `myhome/devices/impl`, `myhome/temperature`, `myhome/occupancy`, `myhome/daemon`, `internal/myhome/shelly/{script,sswitch}`, `internal/myhome/ui/*`, `myhome/ctl/**`.
- **Docs**: CLAUDE.md's "RPC System" and AGENTS.md's "Adding New RPC Methods" updated — the method-registration ritual drops from 4 steps to 2 (no more separate `signatures` map entry or manual type check).

## Verification

- `make generate && make build && make test` (incl. `check-boundaries`) ✅
- `make lint`: 0 issues ✅
- `make test-race`: root + pkg/shelly + pkg/sfr pass, including `internal/myhome` itself (10.9s) — the #356 tests and the new slow-handler test hold under `-race`; `pkg/beem` skip is pre-existing (#374), unrelated
- Wire-format check: `Params`/`Result` moving `any` → `json.RawMessage` doesn't change the MQTT JSON payload (RawMessage marshals its bytes verbatim); no RPC envelope is persisted in `myhome.db`

## Flagged for reviewer judgment

1. **`Register` double-registration behavior**: does *not* panic on re-registering a verb (I initially added a panic matching `pkg/shelly`'s sibling `Registrar`, then reverted — `myhome/occupancy/rpc_test.go` relies on re-registering a verb to swap in a test handler, and there's no `Unregister` API). This matches the *actual* old behavior (`RegisterMethodHandler` only ever panicked on an *unknown* verb, a check that no longer applies without the `signatures` map). Added `myhome.RestoreMethod(verb, *Method)` for test cleanup. A stricter panic-on-duplicate + real `Unregister` is a viable follow-up if preferred, but would need to touch the occupancy test helper too.
2. **`Call[P, R]` runtime type-switch**: `Client.CallE` intentionally stays untyped (`(any, error)`) to keep satisfying the `Client` interface across the real wire client, `FakeClient`, and `DeviceManager`'s in-process short-circuit. The generic `Call[P, R]` wrapper distinguishes three cases at runtime: already-typed `R` (test doubles / in-process), `json.RawMessage` (real wire client, single unmarshal), and a defensive marshal-then-unmarshal fallback. Not explicitly specified by the issue but follows directly from its own `Call[P, R]` sketch.

## Notes

- Base branch has since moved (two main-merge commits, `campaign/main-merge-2`) — no rebase performed yet; will resolve at merge time if needed (the moved commits only touch go.mod/go.sum, pool-pump.js, and linux packaging files, none of which this PR touches).
- Pre-existing, out-of-scope: `internal/myhome/ui/rpc.go`'s `DeviceLookup`/`DeviceListByRoom` UI-conversion code appears to have never matched its actual result types even before this change — left untouched.

## Commits

- rpc: generics-based method registry + json.RawMessage wire types (#361) ([e7ac764](https://github.com/asnowfix/home-automation/commit/e7ac7640d6653a741ecbb4261c824fc04b6aeca4))
- rpc: migrate all method registrations and call sites to Register/Call[P,R] (#361) ([f721e99](https://github.com/asnowfix/home-automation/commit/f721e99ff57e4af72ba2eb6ee151439770b9baa0))
- docs: update RPC method registration steps for Register/Call[P,R] (#361) ([d07b9d2](https://github.com/asnowfix/home-automation/commit/d07b9d23594d1e7ea5b7a485e2436bfe08140487))
- docs(rpc): fix Call[P,R] example in proxy.go doc comment ([af859d9](https://github.com/asnowfix/home-automation/commit/af859d9c8e8610c1ef6b3c78eb3ccf55fb41fe68))

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -933,22 +933,7 @@ Follow this pattern (see temperature and occupancy services as examples):
    - `internal/myhome/occupancy.go` - Occupancy RPC types
    - `internal/myhome/yourservice.go` - Your service RPC types
 
-3. **Add method signature to `internal/myhome/methods.go`:**
-   ```go
-   var signatures map[Verb]MethodSignature = map[Verb]MethodSignature{
-       // ... existing methods
-       YourNewMethod: {
-           NewParams: func() any {
-               return &YourServiceParams{}
-           },
-           NewResult: func() any {
-               return &YourServiceResult{}
-           },
-       },
-   }
-   ```
-
-4. **Create handler in your service package (e.g., `myhome/yourservice/methods.go`):**
+3. **Create handler in your service package (e.g., `myhome/yourservice/methods.go`) and register it — no separate signature-map entry needed:**
    ```go
    type MethodHandlers struct {
        service *Service
@@ -963,22 +948,30 @@ Follow this pattern (see temperature and occupancy services as examples):
    }
    
    func (h *MethodHandlers) RegisterHandlers() {
-       myhome.RegisterMethodHandler(myhome.YourNewMethod, h.handleMethod)
+       myhome.Register(myhome.YourNewMethod, h.handleMethod)
        h.log.Info("Your service RPC handlers registered")
    }
    
-   func (h *MethodHandlers) handleMethod(params any) (any, error) {
-       p, ok := params.(*myhome.YourServiceParams)
-       if !ok {
-           return nil, fmt.Errorf("invalid params type")
-       }
-       
+   // handleMethod's signature IS the method's contract: myhome.Register infers
+   // the request/response wire types from it at compile time, so there is no
+   // separate signature-map entry to keep in sync and no `params any` type
+   // assertion to write — the compiler already guarantees params arrives as
+   // *myhome.YourServiceParams.
+   func (h *MethodHandlers) handleMethod(ctx context.Context, params *myhome.YourServiceParams) (*myhome.YourServiceResult, error) {
        // Your logic here
        return &myhome.YourServiceResult{Data: "result"}, nil
    }
    ```
 
-5. **Register in `myhome/daemon/daemon.go` after device manager starts:**
+   Callers on the other end use the generic `myhome.Call[P, R]` helper to get the same compile-time typing back:
+   ```go
+   result, err := myhome.Call[*myhome.YourServiceParams, *myhome.YourServiceResult](
+       ctx, myhome.TheClient, myhome.YourNewMethod, &myhome.YourServiceParams{Field: "x"},
+   )
+   ```
+   (`myhome.TheClient.CallE` itself stays untyped — it must satisfy the `Client` interface across the wire client, `FakeClient`, and `DeviceManager`'s in-process short-circuit alike — so always call through `myhome.Call`, not `CallE` directly, from application code.)
+
+4. **Register in `myhome/daemon/daemon.go` after device manager starts:**
    ```go
    // Register Your Service RPC methods if enabled
    if options.Flags.EnableYourService {
@@ -1015,7 +1008,7 @@ func NewRPCService(ctx context.Context) (*RPCService, error) {
 ```go
 // CORRECT - Do this!
 func (h *MethodHandlers) RegisterHandlers() {
-    myhome.RegisterMethodHandler(myhome.YourMethod, h.handleMethod)
+    myhome.Register(myhome.YourMethod, h.handleMethod)
 }
 ```
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -76,11 +76,13 @@ No business logic in `myhome/ctl/`. No MyHome-specific code in `pkg/shelly/`. Ut
 
 ### RPC System
 
-All methods share one MQTT topic (`myhome/rpc`). Adding a method requires four steps in order:
+All methods share one MQTT topic (`myhome/rpc`). Adding a method requires two steps in order:
 1. Add `Verb` constant → `internal/myhome/const.go`
-2. Add request/response types → `internal/myhome/<service>.go`
-3. Add to `signatures` map → `internal/myhome/methods.go`
-4. Register via `myhome.RegisterMethodHandler()` — never create a separate MQTT subscription
+2. Add request/response types (if new) → `internal/myhome/<service>.go`, then register the handler via `myhome.Register(verb, handler)` — never create a separate MQTT subscription
+
+`Register[P, R any](verb Verb, h func(ctx context.Context, p P) (R, error))` (`internal/myhome/methods.go`) replaces the old `signatures` map + `RegisterMethodHandler` pair: P and R are normally inferred from `h`'s own signature, so most registrations are a one-line call passing a method reference directly, e.g. `myhome.Register(myhome.HeaterGetConfig, s.HandleGetConfig)`. The wire request's `Params` (and the response's `Result`) travel as `json.RawMessage` end to end and are decoded exactly once, directly into `P` — there is no more runtime `reflect.TypeOf` check and no marshal-then-unmarshal round trip. Registering the same verb twice overwrites the previous registration (there is no `Unregister`); tests that need to swap a handler save the previous `*Method` via `myhome.Methods(verb)` and restore it with `myhome.RestoreMethod` in `t.Cleanup`.
+
+On the caller side, `myhome.TheClient.CallE(ctx, verb, params)` stays untyped (it must satisfy the `Client` interface across the wire client, `FakeClient`, and `DeviceManager`'s in-process short-circuit alike) — use the generic wrapper instead: `result, err := myhome.Call[ParamsType, ResultType](ctx, myhome.TheClient, verb, params)`. For a handler with no meaningful params or result, use `any` for that type parameter (e.g. `myhome.Call[any, *myhome.PoolGetStatusResult](ctx, myhome.TheClient, myhome.PoolGetStatus, nil)`).
 
 ### Key Packages
 
@@ -111,7 +113,7 @@ was present for the discussion that led to filing it.
 
 - **CLI output**: `fmt.Printf()` for user-facing messages; `hlog` for internal/debug logging. Never `log.Info()` in CLI commands.
 - **Config options**: adding any new option requires updating 4 files — `options.go`, `run.go`, `docs/configuration.md`, `myhome-example.yaml`. Env var pattern: `MYHOME_<SECTION>_<KEY>`.
-- **RPC handler tests**: tests that call `myhome.RegisterMethodHandler()` must restore state in `t.Cleanup` and must not call `t.Parallel()` (shared package-level map).
+- **RPC handler tests**: tests that call `myhome.Register()` must restore state in `t.Cleanup` (via `myhome.RestoreMethod`, since there is no `Unregister`) and must not call `t.Parallel()` (shared package-level map).
 - **Database migrations**: Use `COUNT(*)` (returns int) not bool when checking SQLite column existence. See AGENTS.md "Database Patterns".
 - **SQLite database paths**: new databases use a plain relative filename (e.g. `"foo.db"`), matching `myhome.db`. Do not invent a new default directory (e.g. `~/.myhome/`, XDG paths) unless all existing databases already use it. If a flag or config key lets the user supply an absolute path, the `NewStorage` constructor must call `os.MkdirAll(filepath.Dir(path), 0o755)` before opening the file — SQLite cannot create missing parent directories.
 - **File moves**: always `git mv`, never delete-and-recreate (preserves `git log --follow` history).
--- a/internal/myhome/client.go
+++ b/internal/myhome/client.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
-	"reflect"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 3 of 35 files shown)


asnowfix (commented):
## verdict: APPROVE

(gh is authenticated as the PR author, so this is a COMMENT review carrying the explicit verdict. This touches core RPC infrastructure every future method depends on, so I gave it the same depth as #395 — re-derived every claim in a scratch worktree rather than trusting the PR body, including writing throwaway proof-tests for the two highest-risk claims.)

### 1. #356 concurrency correctness — preserved, and the new dispatch surface checked specifically

`client.go`'s design is intact and unchanged in substance: `pendingLock`-guarded `pending` map keyed by `Dialog.Id`, single-reader `dispatch` goroutine, pending-channel registered *before* publish (so dispatch can't race ahead of registration), and `hc.timeout` bounding publish+wait as one budget — all exactly as #356 left it.

The genuinely new surface is `server.go`'s bounded per-message goroutine dispatch (`maxConcurrentRPCs = 16`, a semaphore-gated `go func` per message instead of strictly sequential handling). I read it specifically for the bug classes #356 fixed:
- **Single reader preserved**: `sp.loop` is still the only goroutine reading `sp.from`; the semaphore acquisition happens *before* spawning, so the read loop itself is never concurrent with itself.
- **No head-of-line blocking beyond the bound**: after 16 in-flight handlers, the 17th message legitimately waits (correct, bounded backpressure — not a regression from the unbounded-goroutine risk the issue explicitly warned against).
- **Response correlation unaffected**: each spawned goroutine calls `mc.Publish` independently with its own `response.Id`; nothing shared is written per-request.
- **`TestServer_SlowHandlerDoesNotBlockConcurrentRPC` is a real test, not a shallow one** — I read it and ran it myself. It doesn't just fire two slow calls; it synchronizes on a dedicated `slowStarted` channel to *prove* the slow handler is actually blocked mid-execution before issuing the fast request, then asserts the fast response is published within a tight 200 ms deadline (via `waitPublished`) — a deadline that would only be met if dispatch is genuinely concurrent, and would `t.Fatal` otherwise. I ran it `-race -count=20`: 20/20 pass. I also ran the full `internal/myhome` suite under `-race`: all 16 tests pass, and the `t.Parallel()`-forbidden / `t.Cleanup`-restore convention documented in CLAUDE.md is intact for the migrated tests.

**One thing I traced further and want on record, not as a blocker**: the shared package-level `methods` map is still a naked, unsynchronized `map[Verb]*Method` (same as the old `signatures` map). I confirmed daemon.go still calls `NewServerE` (which immediately starts the dispatch goroutine) *before* several later `Register`/`RegisterHandlers()` calls (EventList, temperature, occupancy, pool — same ordering as the pre-#400 code, verified against 05a1288 byte-for-byte). This is a theoretical, pre-existing window where a very early real RPC request could race a late `Register()` call on that same naked map (Go's runtime can fatal on concurrent map read/write, not just misbehave). **Confirmed unchanged by this PR** — identical ordering exists in the base tree, and it isn't made more likely by bounded-vs-sequential dispatch (the racing party is the startup goroutine, not the dispatch concurrency model). Worth a future issue (e.g. a `sync.RWMutex` around `methods`, or registering everything before `NewServerE` runs), but out of scope here and not something #400 introduced or worsened.

### 2. Wire-format compatibility — verified empirically, not trusted

I wrote and ran a throwaway test (`internal/myhome` package, deleted after use) that: (a) marshaled a `response{Result: *any}` the *old* way and a `response{Result: json.RawMessage}` the *new* way for the same real result type (`DeviceSetRoomParams`), and confirmed **byte-identical wire output**; (b) fed the *old*-shaped bytes into the *new* `response` struct and confirmed it decodes correctly end-to-end into the target type. This also proves the specific cross-version property the issue calls out ("CLI and daemon must still be release-compatible across versions") — an old server's bytes decode cleanly through a new client.

On the "no envelope persisted in `myhome.db`" claim: `request`/`response` are unexported (lowercase) types in `internal/myhome/proxy.go`, so it's not just true today but structurally impossible for `myhome/storage` (which only imports the package's exported `myhome.Device` etc.) to persist them. Confirmed by inspection, not just grep.

### 3. The two flagged design judgment calls

**(a) Silent overwrite on double-registration.** I agree with treating this as a real (if narrow) risk, not a rubber-stamp. The concrete failure mode: two services accidentally both claiming the same `Verb` constant (bad merge, copy-paste) would silently leave one dead with no signal at startup, versus a loud, cheap-to-diagnose panic at exactly the moment (process start, before any live traffic) where it's easiest to catch. I checked whether this has *already* happened anywhere in the current ~70 registrations (grepped for duplicate verb registration) — it hasn't; the one apparent hit (`DeviceShow` appearing twice) was a doc-comment example, not a real second call site. So this is a **latent design gap, not an active bug**. A workable fix that wouldn't need to touch the occupancy test's actual swap call: gate the overwrite check on `testing.Testing()` (stdlib, Go 1.21+, available here since go.mod pins 1.25) — panic on duplicate registration when *not* running under `go test`, silently overwrite when running under it. That keeps `myhome/occupancy/rpc_test.go`'s `Register(verb, h)`-to-swap pattern working unchanged while closing the production gap. Worth a fast follow-up issue; not a blocker for this PR, since the current behavior is a strict superset of the old behavior (old `RegisterMethodHandler` never checked for duplicates either — this PR doesn't regress anything, it just doesn't yet add the stronger guard the issue's spirit invites).

**(b) The three-way `Call[P, R]` type switch.** I checked whether the `default:` marshal-fallback branch is dead/defensive-only or actually load-bearing. It's load-bearing: I traced `DevicesMatch`/`DeviceLookup`'s registered handlers (return `*[]devices.Device`, an interface slice) against `client.go`'s `LookupDevices` and `myhome/ctl/room/list.go`'s call sites (both use `Call[..., []myhome.DeviceSummary]`, a narrower concrete struct). For the **wire client**, `CallE` always returns `json.RawMessage`, so this always hits the RawMessage case, never the fallback — no risk there. For `DeviceManager`'s **in-process short-circuit** (bypassing MQTT via `Method.Call` directly), `out`'s dynamic type is the handler's literal `R` (`*[]devices.Device`), which doesn't equal the caller's chosen `[]myhome.DeviceSummary` — so it *does* fall to `default:`, re-marshals, and unmarshals into the narrower type. This works today because `myhome.Device` embeds `DeviceSummary` (same JSON tags), so the JSON shapes are genuinely compatible — this is a deliberate "richer domain object server-side, narrower view client-side" pattern that already existed pre-#400 (the old client did its own marshal→unmarshal round trip for exactly this reason). So: the fallback doesn't hide a *new* bug class — the underlying "caller and handler must independently agree their types are JSON-compatible, with nothing enforcing it" risk is exactly as old as the JSON-RPC design itself, not introduced or widened by `Call[P, R]`. I'd suggest, as a non-blocking style note, a doc comment on the `default:` case making explicit that it's real (not just paranoid) and specifically for the in-process-short-circuit-with-a-narrower-caller-type pattern — future readers hitting it in a debugger will otherwise wonder if it's dead code.

### 4. Mechanical migration — spot-checked, no defects found

Zero residual `TheClient.CallE()` call sites anywhere (confirmed by grep across the whole tree). The `RegisterMethodHandler`/`.CallE(` hits that do remain are all `pkg/shelly`'s unrelated sibling `Registrar`/`Device.CallE` system (a different layer per the Three-Tier Layer Rule, correctly untouched). Spot-checked ~15 of the ~37 `Call[P, R]` instantiations across `myhome/ctl/{room,db,events,temperature,heater,shelly,pool,show}` against their registered handlers' actual `P`/`R` — all type-correct or, where they intentionally differ (the `DeviceSummary` case above), verified JSON-compatible by design, not by accident.

### 5. CI — confirmed on the live PR, one unrelated failure identified and diagnosed

`build`, `lint`, `race`, `coverage-delta` all pass on the current head. `build-pr-deb` fails, but I traced it to the exact root cause and it has nothing to do with this PR's code: `dpkg-deb: error: ... 'Version' field value 'campaign/main-merge-2-5-ge50f859': version number does not start with digit`. The repo has a series of campaign checkpoint tags (`campaign/355-ci-hardening`, `campaign/359-module-collapse`, `campaign/main-merge-1`, `campaign/main-merge-2`, etc.) that `git describe --tags` picks up as the "nearest tag" for Debian versioning — since these tags don't start with a digit, any build whose nearest reachable tag is one of them fails `dpkg-deb`'s version parsing, regardless of what code changed. Reproduced locally: `git describe --tags` on this exact branch returns `campaign/359-module-collapse-4-gaf859d9`. This is a CI/tagging infrastructure defect (the campaign's own tagging convention colliding with a build script's `git describe` assumption), pre-existing and unrelated to RPC code — worth a follow-up issue (e.g. `git describe --tags --match 'v*'` to restrict matching to real version tags), not a blocker here. `build`/`lint`/`race`/`coverage-delta` are the substantive gates and they're clean.

### 6. `internal/myhome/ui/rpc.go`'s DeviceLookup/DeviceListByRoom conversion — confirmed genuinely pre-existing

Diffed this file against 05a1288 byte-for-byte outside the params-decoding preamble (which changed mechanically to match the new `Action` signature). The `res.([]myhome.Device)` type assertion at line 55 is unchanged and was *already* dead on arrival before this PR: `DeviceLookup` and `DeviceListByRoom` actually return `*[]devices.Device` and `*myhome.DeviceListByRoomResult` respectively (neither matches `[]myhome.Device`), so the UI-conversion branch has never fired for either verb, in old or new code. Correctly left untouched by this PR — worth its own follow-up issue (the UI likely renders these two RPC results without the nicer `DeviceView` transform), but not something #400 should have absorbed while touching adjacent code.

### Summary

Core rewrite is sound: the #356 concurrency guarantees survive semantically (not just syntactically), the new bounded-dispatch surface doesn't reintroduce the bug classes #356 fixed, wire compatibility is proven byte-for-byte including cross-version decode, and the mechanical migration is complete and spot-check-clean. Both flagged judgment calls are reasonable as shipped, with concrete, low-cost follow-ups identified for each. Two out-of-scope findings (the naked `methods` map's startup-window race, and the `campaign/*` git-tag versioning defect) are for the orchestrator, not blockers here.

Issue #408, opened 2026-07-28
internal/myhome: naked package-level methods map has a theoretical startup-ordering race


asnowfix:
## Problem

`internal/myhome`'s method registry (`methods` map, written to by every `Register` call across the codebase's `init()`/setup functions, read by the RPC dispatch loop) is an unsynchronized package-level map. If the daemon's RPC dispatch loop starts consuming requests before every package's `Register` calls have completed, a concurrent read (dispatch) and write (a not-yet-finished registration) on the same map is a data race.

Confirmed **pre-existing**: identical ordering exists on the base tree before #361 (PR #400) — #361's generics-based registry rewrite did not introduce or worsen this, it inherited the same map-lifecycle assumption from the original `signatures` map design. Found during the #361 PR review (master reviewer agent), which verified this by diffing daemon.go's startup sequence against the pre-#361 tree.

## Why it hasn't bitten in practice

In today's actual daemon startup path, all `Register` calls happen synchronously during package `init()` / `NewDaemon` setup, strictly before `daemon.Run()` starts the MQTT subscription that feeds the dispatch loop — so there's no real concurrent access today. The race is latent, not live.

## Suggested fix

Either: (a) guard the `methods` map with a `sync.RWMutex` (cheap, matches the pattern already used for `pendingLock` in `internal/myhome/client.go` since #356), or (b) make the ordering an explicit, enforced invariant (e.g. a `sync.Once`-guarded "registry sealed" flag that dispatch checks) so a future refactor that runs registration concurrently with dispatch fails loudly instead of racing silently.

## Suggested owner

Natural fit for #364 (goroutine lifecycle / daemon startup error propagation) since it's in the same "daemon startup ordering" family, or a small standalone fix — either is reasonable.

Part of #368. Found during review of PR #400 (#361).

PR #409, opened 2026-07-28
arch: dependency-injection pass — remove package-level singletons (#362)

### LINKED ISSUE #362: arch: dependency-injection pass — remove package-level singletons (TheClient, options.Flags in libraries, MQTT globals)

asnowfix:
Closes #362. Part of umbrella #368 (Wave 3, last item).

## What

Six commits removing package-level mutable singletons, replaced with explicit constructor/parameter injection (no DI framework — plain Go, per CLAUDE.md's simplicity-over-generality rule):

1. `panicOnBugs` passed explicitly to `ui.Start`/`proxy.Handle` instead of reading `internal/global.PanicOnBugs`.
2. (test-build fix for #1)
3. **RPC method registry**: package-level `var methods map[Verb]*Method` — previously **unsynchronized, a real latent race** (related to #408, filed during the #361 review) — replaced by `myhome.Registry`, injected explicitly into every service constructor and reader. RPC-handler tests now build their own `Registry` and run with `t.Parallel()`. CLAUDE.md updated.
4. **`myhome.TheClient` removed entirely** (~67 files) — the biggest, riskiest single change here. CLI composition root (`ctl.go`) stores the client via `myhome.NewContextWithClient` on `cmd.Context()`; library helpers (`Foreach`, `ConfigureDevice`, `blu.ResolveMac`) take it as an explicit parameter. The MCP server handlers needed special-casing since `mcp-go`'s `ServeStdio` context doesn't carry `cmd.Context()`'s value.
5. **The "worst single edge"** (issue's own words): `myhome/mqtt` no longer imports `options`; `GetClientE`'s old fallback (which silently built an under-configured client) is gone — construction now happens explicitly in `NewClientE`. `internal/myhome/net.MyResolver` and `myhome/daemon/watch.StartMqttWatcher` decoupled the same way.
6. **`pkg/shelly/mqtt`'s `SetClient`/`GetClient`/`ResetClient`** global registry replaced with the same context-based pattern; ~19 test call sites converted.

## Verification

`make generate && make build && make test && make lint && make test-race && make check-boundaries` — all green. `test-race` excludes only the pre-tracked `RACE_SKIP_PACKAGES` (#372/#373/#374), none touched by this PR.

## Acceptance criteria status (partial — see below)

- `TheClient` grep-clean except a doc comment and one test function name referencing the historical symbol for context.
- `myhome/mqtt`, `internal/*`, `pkg/*` no longer import `myhome/ctl/options`.
- **`myhome/daemon` is partial**: `myhome/daemon/watch` is clean, but `daemon.go`/`run.go` (~80 `options.Flags.*` reads across every daemon feature — the daemon's actual startup path controlling live house hardware) were judged too large/risky to transcribe in this pass and are **not done**. Flagged as a dedicated follow-up — please advise whether that should be a new issue or reopen scope on #362.
- `SetClient`/`ResetClient`/`GetClient(ctx)`-style accessors gone in both `internal/myhome` and `pkg/shelly/mqtt`.
- RPC handler tests run with `t.Parallel()`.

## Out-of-scope findings (not fixed, flagged on #362)

`hlog.Logger` (hundreds of call sites), `pkg/shelly/device.go`'s `deviceMqttRegistry` (needs `Device` to hold a client reference it doesn't currently have), and two globals not in the issue's original inventory: `pkg/shelly`'s `registrar`, `pkg/shelly/mqtt`'s `mqttChannel`.

## Flagged for reviewer judgment

- **The `TheClient` removal breadth** (~67 files) is exactly the kind of change worth a careful second read given its size, even though every build/test/race/lint check passes.
- **Three incidental correctness fixes bundled into the DI commits** (documented in their own commit messages, not separated out): the registry mutex, the `GetClientE` under-configured-fallback removal, and a watchdog-ticker-zero clamp. None are behavior regressions, but they're "found this while moving the code" fixes riding along with a refactor — worth confirming each is genuinely correct and not accidentally scope creep.
- **`myhome/daemon`'s incomplete DI** — is leaving `daemon.go`/`run.go` on `options.Flags` acceptable for this PR, or should it block merge?

Implemented by the issue agent (549k tokens, 463 tool uses — a large pass). Verified locally by the agent; not yet independently re-verified by the coordinator given the size — flagging that explicitly for the reviewer.

## Commits

- di(#362): pass panicOnBugs explicitly instead of reading global.PanicOnBugs ([4190ca6](https://github.com/asnowfix/home-automation/commit/4190ca64c11214e70713b2104869dbfe80bc31ea))
- test(#362): fix reverse_test.go build after Handle's panicOnBugs param ([6f2322d](https://github.com/asnowfix/home-automation/commit/6f2322d1b47602a84b5d6720cf771ea5565ad734))
- di(#362): make the RPC method registry an injected *Registry ([d9a4ff1](https://github.com/asnowfix/home-automation/commit/d9a4ff122755c2bc5670ce0e51d2e35fab0ddd4c))
- di(#362): remove myhome.TheClient package-level global ([93926ef](https://github.com/asnowfix/home-automation/commit/93926ef1787927fdc2cfb928311a25c901a421b9))
- di(#362): decouple myhome/mqtt and internal/myhome/net from myhome/ctl/options ([df39455](https://github.com/asnowfix/home-automation/commit/df3945574ad7b4836c484ca4ce2378cdf231831c))
- di(#362): replace pkg/shelly/mqtt's SetClient/GetClient/ResetClient global registry ([aaade10](https://github.com/asnowfix/home-automation/commit/aaade100b9a3865996269bfccee7c9c096bee82a))


asnowfix:
**This PR is the fix for #562**, and it has been sitting unreviewed since 2026-07-28.

#562 was filed today after PR #561 recovered only 13.6% of `internal/shelly/scripts`' runtime and
could not touch the remaining **86%**: `pkg/shelly/mqtt` holds the client in a process-global and
`SetClient()` panics on a second distinct client (`client.go:44`), which nearly every test in the
three `pool_pump*` files swaps via `ResetClient()` / `SetClient()`. Adding `t.Parallel()` there risks
tests silently running against each other's mock clients — a green suite proving nothing.

This PR deletes exactly that construct and replaces it with context-based injection. The maintainer
has chosen that approach for #562, so reviving this is the work rather than writing it again.

Context accumulated since this was opened, which makes it more valuable than when it was filed:

- `internal/shelly/scripts` grew **507 s → 686 s in four days** and crossed two independent
  600-second ceilings — Go's default `-timeout` **and** the agent harness's cap on a foreground
  command, so any agent running `make test` in the foreground is now killed mid-run. Tracked as #552;
  the timeout was raised three times as a stopgap (#554, #556, #558) and lowered once by #561.
- Four device-script defects this month passed CI and failed on hardware, so test *trustworthiness*
  is the scarce resource. That is also an argument for this PR's other half — the unsynchronised RPC
  method registry it describes as "a real latent race".

**Being rebased now.** `main` has taken roughly fifteen merges since this was opened, several touching
the same test packages, so the first question is whether it still applies. Review follows the rebase.

asnowfix:
## Triage 2026-09-06 — needs a **maintainer decision**, and has a structural problem no rebase fixes

Reviewed while clearing the `stale-unreviewed` backlog.

**This PR does not target `main`.** Its base is `worktree-feature+golang-expert-review`
(`f9156713`) — itself a stale feature branch, not the default branch. So its 97 files / +1101/−587 are
measured against a base that is not going anywhere, and "MERGEABLE" here means mergeable *into that
branch*, which merges nothing into the product. That is the first thing to settle, and it is a decision
rather than a task: **retarget onto `main`, or close.**

**The underlying issue #362 is still open and still real** — package-level singletons (`TheClient`,
`options.Flags` in libraries, MQTT globals). Nothing in campaign #401 addressed it. There is prior art
in the tree now that did not exist in July: #562 removed `pkg/shelly/mqtt`'s process-global client
because it gated 86% of the slowest test package. That is the same class of problem this PR attacks, and
it landed on its own — evidence the work is tractable in slices.

**Recommendation:** do not rebase 97 files wholesale onto seven weeks of drift. Either
- **retarget to `main` and split** into reviewable slices, taking the highest-value singleton first
  (#562's precedent suggests the payoff shows up as test parallelism, which is currently the suite's
  binding cost at ~612s for one package); or
- **close it** and keep #362 open with a note that the branch exists and can be mined for the mapping
  work already done.

Either way this should not sit as-is: as it stands it is invisible to `main` and cannot be merged into
anything that ships.

Issue #406, opened 2026-07-28
chore(daemon): remove legacy SolarAutomation state machine and pool.solar.* daemon config


asnowfix:
Part of #401. Depends on: #402, #403, #405 (must land and be verified working on real hardware
first — see rationale below). Related: #404 (independent, no ordering constraint with this issue).

## Background

This is the final cleanup step of the solar-automation redesign (see #401). It must only be merged
**after** #402 (on-device runtime accumulator), #403 (daemon solar-aggregator publish), and #405
(on-device solar hysteresis) are deployed and verified working against real hardware — since the
code being deleted here is currently non-functional anyway (see #401's root-cause section: the
`readPoolKVSFloat` bug on `script/pool-pump/speed` means solar automation never actually starts
today), this carries low regression risk, but sequencing it last avoids a window where neither the
old nor the new mechanism is in place if something in the new implementation needs a quick follow-up
fix.

### Delete entirely

- `myhome/daemon/solar_automation.go` — the whole file: `SolarConfig`, `PumpController`,
  `RuntimeTracker` interfaces, `SolarAutomation` struct and its `step()`/`run()` state machine,
  `shellyPumpController` and its MQTT status-subscription/retry/confirm machinery, and
  `computeRuntimeTargets`/`readPoolKVSFloat`/the associated `kvsKey*` consts (unless
  `readPoolKVSFloat` was already relocated to `pool_notices.go` by #402, in which case only delete
  the duplicate here).
- `myhome/daemon/solar_automation_test.go`
- `myhome/daemon/solar_notice_test.go`
- `myhome/daemon/pool_runtime_tracker.go` — confirm no remaining caller first (#402 should have
  already removed `PoolNotices`'s dependency on it; grep for `PoolRuntimeTracker`/
  `NewPoolRuntimeTracker` across `myhome/daemon/` before deleting).
- `myhome/daemon/pool_runtime_tracker_test.go`
- The `if options.Flags.PoolSolarEnabled && ...` daemon-wiring block that constructs
  `shellyPumpController`/`SolarAutomation` (in `myhome/daemon/run.go` or `daemon.go`, wherever the
  Beem watcher and pool tracker are wired up).

### Delete config surface

`myhome/ctl/options/options.go`: delete `PoolSolarEnabled`, `PoolSolarStartThresholdW`,
`PoolSolarStopThresholdW`, `PoolSolarStartDelay`, `PoolSolarStopDelay`, `PoolSolarMinVolumeTurnover`,
[...]

Issue #402, opened 2026-07-28
fix(pool-pump): track today's cumulative pump runtime/turnover on-device (fixes ctl pool status)


asnowfix:
Part of #401.

## Background

`myhome/daemon/pool_notices.go`'s `PoolNotices.ComputeTurnover()` — used by both the
`pool.turnover_today` notice and the `pool.getstatus` RPC (i.e. `ctl pool status` and the web UI) —
reads the pool device's KVS key `script/pool-pump/speed` and parses it with
`fmt.Sscanf(v, "%f", &num)` to compute a flow rate. That key actually holds a semantic string
(`"eco"`/`"mid"`/`"high"`/`"max"`), set by `pool-pump.js`'s `CONFIG_SCHEMA.preferredSpeed` (default
`"eco"`) — never a numeric RPM. The parse always fails, so `ComputeTurnover()` **always returns an
error today** — `ctl pool status` and the turnover notice are currently broken, unconditionally.

The device itself already computes the correct flow rate correctly, via
`internal/shelly/scripts/pool-pump.js`'s `computeFlowRate()`, which maps the speed string through
`CONFIG.ecoRpm`/`midRpm`/`highRpm`/`maxRpm`. The fix: have the script itself compute and persist
today's cumulative runtime and turnover-achieved to new KVS keys, and have the Go side simply read
those pre-computed values instead of re-deriving flow rate in Go (where the string→RPM mapping isn't
available without duplicating JS logic).

This also lays groundwork for a separate, related redesign (tracked in a sibling issue, part of the
same umbrella #401) that needs the device to know its own daily runtime independent of the daemon's
`events.db`.

### `internal/shelly/scripts/pool-pump.js`

1. Add two new keys to `STORAGE_KEYS` (the `Script.storage`-based synchronous-at-boot state,
   alongside the existing `forecastUrl`/`scheduleMode`):
   - `runtimeSec` → storage key `"runtime-sec"` — cumulative pump-on seconds today (completed
     intervals only; open intervals are checkpointed periodically, see below).
   - `runtimeDate` → storage key `"runtime-date"` — the `"YYYY-M-D"` date the above count applies to,
     for day-rollover detection (reuse the existing date-formatting helper already used by
     `shouldRefreshForecast()`).

2. Add new `STATE` fields: `runtimeTodaySec` (number), `runtimeDate` (string), `runStartTs`
   (number|null — `Date.now()` when the current ON interval began; null when off),
[...]

Issue #250, opened 2026-06-07
script emulator: enforce Shelly device resource limits with device-test vs device-extension modes


asnowfix:
## Background

`pkg/shelly/script/run.go` emulates a Shelly device for Go-based script tests. It currently enforces **no resource limits**: timers, event handlers, MQTT subscriptions, and `Shelly.call` can all be created without bound. This means tests could pass while scripts silently violate device limits that will crash or misbehave on real hardware.

The crash fixed in branch `bugfix/too-many-calls-blu-listener` (`blu-listener.js`: "Too many calls in progress") was **invisible in tests** because `Shelly.call` in the emulator invokes callbacks synchronously, so the concurrent-call count never accumulates. The emulator needs to be able to detect such violations.

At the same time, a future use-case requires running scripts that intentionally **exceed** device constraints (e.g. scripts split across devices, or daemon-side script execution where device limits don't apply). These two needs require two distinct modes.

---

### `device-test` mode (default for unit tests)

The emulator mirrors real Shelly device limits. **Any violation causes an immediate test failure** (panic / `t.Fatal`-equivalent) with a message that matches what the real device would emit. The goal is to catch limit violations during CI before deployment.

### `device-extension` mode

No limits are enforced. Intended for running scripts in the daemon that are "too large" for a single physical device — scripts that aggregate state across multiple Shelly devices, run on non-device hardware, or deliberately exceed hardware constraints. State should persist identically to `device-test` mode (same `DeviceState` mechanism).

The mode is set via a field on `DeviceState` (or a constructor flag), e.g.:

```go
deviceState := &script.DeviceState{
    Mode: script.DeviceTestMode,   // or script.DeviceExtensionMode
    KVS:  ...,
    ...
}
```

---

## Limits to enforce (from AGENTS.md and Shelly docs)

All of these are **per-script** on real Shelly Gen2 devices:

| Resource | Limit | Shelly error / behaviour |
|---|---|---|
| Concurrent `Shelly.call` in progress | 5 | `Uncaught Error: Too many calls in progress` |
| `Timer.set` active at once | 5 | silently fails / undefined behaviour |
| `Shelly.addEventHandler` registrations | 5 | silently ignored above limit |
| `Shelly.addStatusHandler` registrations | 5 | silently ignored above limit |
| `MQTT.subscribe` topics | 10 | silently ignored above limit |
| HTTP registered endpoints | 5 | silently ignored above limit |
[...]

PR #426, opened 2026-08-04
fix(pool-pump): bound in-flight RPCs, catch up on restart, and fit the device heap (#421)

### LINKED ISSUE #421: fix(pool-pump): "Too many calls in progress" crash kills script + no catch-up after restart
Part of #401 (found during that issue's live-device verification, but an independent regression —
not one of #401's ordered sub-issues).

## Incident

On 2026-08-03, live-verifying #402/#403/#404/#405 against the real pool pump device
(`filtration-hiver`, id `shellypro1-ec62608c0230`, a Shelly Pro1) after uploading the current `main`
build of `pool-pump.js` (minified — the unminified build doesn't fit the device's script storage,
separate finding, see `docs/pool-pump.md` follow-up), the pump was found running long past its
scheduled 18:14 evening-stop time with no way to turn it off via schedule. Root cause: the script
had crashed hours earlier and stayed crashed — `Script.GetStatus` for id 2 returned:

```
"running": false,
"error_msg": "Uncaught Error: Too many calls in progress\n
 at ...ONFIG_KEY_PREFIX+e,value:n})\n
in function \"storeValue\" called from storeValue(\"turnover-today\",computeTurnoverToday(t))\n
in function \"e\" called from ...TASK_INDEX];TASK_INDEX++,e()\n
in function called from system"
```

Restarting the script (`Script.Stop` + `Script.Start`) brought it back up, but it **crashed again
within seconds**, this time from a different call site:

```
"error_msg": "Uncaught Error: Too many calls in progress\n
in function \"storeValue\" called from storeValue(STATE_KEYS.activeOutput,STATE.activeOutput)\n
in function \"e\" called from ...TASK_INDEX];TASK_INDEX++,e()"
```

Two distinct call sites (`persistRuntimeState`'s turnover mirror, and `saveState`'s active-output
mirror) hit the same crash class — this is systemic to how `queueTask`/`storeValue` interact, not a
one-off. The pump was recovered by manually confirming `Switch.Set {id:0, on:false}` (it turned out
the crashed run of `handleEveningStop()` had actually already turned the switch off before crashing
on its own follow-up KVS write — the manual call was a no-op safety confirmation, not what actually
stopped the pump).

## Bug A — `queueTask`/`storeValue` crash on "Too many calls in progress"

**Root cause**: `processTaskQueue` (`internal/shelly/scripts/pool-pump.js:429-446`) dispatches one
queued task function per 200ms timer tick, which only serializes *when the JS task functions run* —
it does not track or rate-limit the actual asynchronous `Shelly.call` RPC invocations those tasks
trigger. `storeValue` (`pool-pump.js:567-580`) is deliberately fire-and-forget (no completion
callback — "to avoid callback depth issues"), so nothing in the script ever knows how many of its
own previous `Shelly.call`s are still in flight on the device.

Real Shelly Gen2 firmware enforces a hard ceiling of 5 concurrent RPC calls per script (see
`CLAUDE.md` "Per-script limits"). Several independent triggers each fire `Shelly.call` more or less
in parallel — the `TASK_QUEUE`'s own 200ms-tick writes, the 60s `runtimeFlushTimer`
(`flushRuntimeCheckpoint` → `persistRuntimeState`, itself enqueuing 2 more calls), the 30s solar
hysteresis tick (`checkSolarHysteresis`, #405), and — most acutely — the boot-time `continueInit()`
sequence, where `saveState()`, `persistRuntimeState()` (via `enforceOutputState` →
`startRuntimeAccounting`), and the 4-step `initSteps` sequence (`Sys.SetConfig`,
`applyComponentNames`, `verifySchedules`, ...) all queue or fire calls within a very short window.
When enough of these overlap, the device raises `Too many calls in progress` as an uncaught JS
exception from whichever call tips it over 5. Nothing in `processTaskQueue` (or the wider script)
wraps task execution in try/catch, so that single uncaught exception kills the **entire script**, not
just the offending task — every subsequent scheduled/queued operation stops working, including the
`doStop()` a missed-but-still-pending schedule call would otherwise trigger.

The code comment at `pool-pump.js:1026-1031` explicitly names this exact failure mode as one PR #394
already fixed once — this is a recurrence in a different call path #394 didn't cover (init-time /
`saveState` mirror, not just the periodic runtime-checkpoint pair PR #394 addressed).

**Suggested fix direction** (not prescriptive — implementer's judgment): give the task queue (or
`storeValue` itself) a real notion of calls-in-flight — e.g. `storeValue` accepts an optional
completion callback, `processTaskQueue` only advances once the previous task's underlying
`Shelly.call` has actually completed (or a bounded number are in flight), rather than trusting a
fixed 200ms tick to be "enough time." Wrapping `task()` in `processTaskQueue` with a try/catch (log +
continue instead of letting one bad task kill the script) is a second, independent layer of
defense worth adding regardless of the root fix — a future unrelated crash in one queued task
shouldn't be able to kill schedule-driven pump control again.

## Bug B — no catch-up-on-restart against the current schedule window

Independent of Bug A: `enforceOutputState()` (`pool-pump.js:2213-2259`), called on every
(re)start, only mirrors whatever the physical switch's current output state already is into
`STATE.activeOutput` — it never checks whether *now* falls inside or outside today's currently
active schedule window. Morning-start/evening-stop times are computed dynamically
(`computeRunHours`, `makeTimespec`, `updateScheduleMode`, `pool-pump.js:1931-2020`) and written
directly onto the device's own `Schedule` jobs — the script itself has no independent "should I be
running right now" check at boot, only reacts to whichever discrete schedule job fires next.

Consequence: if the script restarts (crash recovery, script re-upload, device reboot) at any point
between the last-passed schedule boundary and the next one, it silently adopts whatever physical
state it finds — on or off — with no correction, even if that state is stale relative to today's
schedule. This is exactly what happened tonight: the crash-recovery restart left the pump on well
past evening-stop, and nothing in the restart path noticed or corrected it.

**Suggested fix direction**: on boot, after `loadState()`/`enforceOutputState()`, compare current
time-of-day against the currently active mode's window (reuse the same `Schedule.List`
enable/timespec data already read in `updateScheduleMode`, or track the last-computed
morning/evening boundaries in `Script.storage` alongside `scheduleMode`) and call `doStart(...)` /
`doStop('Restart catch-up: outside scheduled window')` as needed to bring the physical state back in
line with schedule before falling back to "wait for the next schedule tick."

## Test plan — write the reproducing test FIRST (Prove-It / TDD)

Before implementing either fix, add a test that reproduces Bug A's crash and fails against current
`main`. This currently isn't possible against the existing script-emulator harness
(`pkg/shelly/script/run.go`'s `Shelly.call` — see `run.go:180-209`) because it executes every method
synchronously with **no concurrency ceiling and no artificial async delay at all** — there's no
notion of "calls in flight," so no amount of queueing logic could ever trip a limit that isn't
modeled. This is exactly issue #250's territory (open: "script emulator: enforce Shelly device
resource limits with device-test vs device-extension modes").

**Scope decision for this issue**: implement only the minimal slice of #250 needed to reproduce this
one bug — a per-VM in-flight-call counter plus a configurable artificial completion delay for
`Shelly.call`, enforced only in `DeviceTestMode` (panic/fail at >5 concurrent, matching real
hardware), inert in `DeviceExtensionMode` (daemon-side `myhome/scripthost`, per
`docs/*` / project convention that daemon-side execution must stay unlimited). Do **not** attempt
the rest of #250's scope here (timer/event-handler/status-handler/MQTT-subscription limits,
5/5/5/10 ceilings) — leave that to #250 itself, only cross-reference it as "partially addressed by
this issue's concurrent-call slice."

1. **New emulator capability** (`pkg/shelly/script/run.go`): concurrent-call counter + delay hook,
   `DeviceTestMode` ceiling enforcement as described above.
2. **New failing test** (`internal/shelly/scripts/pool_pump_test.go`): drive a boot sequence (or a
   tight run/stop/run cycle) through `RunWithDeviceState` with the new emulator mode/delay configured
   so enough `Shelly.call`s legitimately overlap to exceed 5 concurrent — assert the script crashes
   with the `Too many calls in progress` message today (proving the reproduction is real), then after
   the Bug A fix lands, assert it no longer crashes and all queued KVS writes still land correctly
   with the schedule/state remaining consistent.
3. **Bug B test**: simulate a restart at a simulated time known to be outside the (mocked) current
   schedule window with the switch left "on" in `DeviceState`; assert the fixed `enforceOutputState()`
   path calls `doStop()` and the switch ends up off, without waiting for a schedule tick.
4. Implement both fixes; confirm both tests now pass, and re-run the full existing
   `pool_pump_test.go` suite to confirm no regressions.
5. **Manual re-verification on real hardware** (`filtration-hiver`, or preferably the spare Shelly
   Plus 1 named `development` — its device-type auto-detection is purely by switch count
   (`pool-pump.js:321-331`), so it takes the same single-output code path with zero script changes;
   point `script/pool-pump/preferred` at its own device ID and use short-interval test `Schedule`
   jobs instead of daily ones to exercise start/stop repeatedly without daily-cadence waiting):
   confirm the script survives a normal boot without crashing, and that a restart mid-window
   correctly self-corrects per Bug B's fix.

## Files touched (expected)

- `internal/shelly/scripts/pool-pump.js`
- `internal/shelly/scripts/pool_pump_test.go`
- `pkg/shelly/script/run.go`
- `pkg/shelly/script/run_test.go` (or equivalent, for the new emulator capability itself)

## Relationship to #401 / #406

This must be fixed and re-verified live before #401's #406 (removal of the legacy
`SolarAutomation` daemon code) can proceed — #406 is explicitly gated on #402/#403/#405 being
verified working against a real device, and tonight's session found the opposite (a crash), so that
gate remains **not met**. See `docs/401-campaign-live-verification-status.md` for the fuller
handover context.

Real-hardware re-verification of this fix (step 5 of the test plan above) is tracked as the first
test case of **#422** ("test: establish real-device integration test campaign using spare Shelly
Plus 1 (development)") — see that issue for the device-availability precondition and the
recording template used for that verification.

asnowfix:
Fixes #421. **Do not merge yet — see "Blocked on real hardware" below.**

## What this fixes

**Bug A — `Too many calls in progress` killed the entire script.** `processTaskQueue` dispatched one
queued task per 200ms tick, which serialises only *when task functions run* — it had no idea how
many `Shelly.call` RPCs those tasks left in flight, and `storeValue()` is fire-and-forget. Real Gen2
firmware caps concurrent RPCs per script at 5; the 6th raises an uncaught exception that kills the
whole script, stopping all schedule-driven pump control. That is how a live pump was found running
hours past its scheduled evening stop.

Fix: track in-flight calls and have `processTaskQueue` skip a tick while `MAX_CALLS_IN_FLIGHT` (4,
one below the device ceiling) are outstanding, leaving `TASK_INDEX` unchanged so the task retries.
Plus a `try/catch` around `task()` as an independent second layer — no single queued task should be
able to kill schedule-driven pump control again, whatever the cause.

**Bug B — no catch-up against the schedule window on restart.** `enforceOutputState()` only mirrored
the physical switch at boot; it never asked whether *now* is inside today's window, so a restart
mid-window silently adopted stale state until the next tick, possibly hours later. Adds
`restartCatchUp()`, which reads the active mode's start/stop jobs from `Schedule.List` (the same
source `updateScheduleMode` writes) and calls `doStart()`/`doStop()` when reality disagrees. Handles
windows wrapping past midnight; skips unresolvable/symbolic timespecs rather than guessing.

**Emulator:** adds a per-VM concurrent-call counter and artificial completion delay, enforced only in
`DeviceTestMode`. This is the minimal slice of #250 needed to reproduce the crash in tests; the rest
of #250's scope is deliberately left to that issue.

**Fail-loudly self-check.** The throttle works by reassigning `Shelly.call` to a wrapper. If that
ever stops being permitted, the counter stays 0, the throttle goes inert, and the crash returns —
*while every emulator test still passes*, because goja allows a reassignment a device might refuse.
Failing open and silently is the dangerous case, so it is asserted, not assumed. Two failure modes
are covered (assignment rejected; assignment took but nothing tracked), both printing a diagnostic
that names the cause, the consequence, an explicit "read issue #421 first", and the alternative fix
direction — so whoever meets this log next does not re-derive the root cause. Both also emit a
`pool.call_tracking_broken` notice event. A healthy boot logs a positive confirmation, so silence is
never ambiguous. Uses `print()` rather than `log()` because `log()` is gated on
`CONFIG.enableLogging` and would hide the message in production. Lines are kept short because the
device truncates UDP debug lines at ~128 chars.

## Verified on real hardware (Shelly Plus 1, fw 1.7.5)

- Wrapper installs on mJS (`patched=true`).
- The 12-concurrent-`Script.Eval` burst that crashed the pre-fix script on its first attempt now
  survives. Throttle engaged 3x at 4 in flight; the `try/catch` never fired — the root cause is
  prevented, not merely caught.
- Queued KVS writes still land (`runtime-sec`, `turnover-today`, `active-output`) — survival is not
  bought by dropping work.
- Restart catch-up emitted `pool.pump_stop {"reason":"Restart catch-up: outside scheduled window"}`
  and physically opened the relay.
- Both self-check failure modes produce the full untruncated diagnostic and the notice event.

`make test`: 47 packages ok, 0 failures.

## Memory work (why the last commit exists)

The first working version could not run on real hardware at all — `out_of_memory` on the production
Pro1 and on a Plus 1 configured to match it, stranding the pump ON. Measurements (total JS heap
23030 bytes, shared across all scripts, `watchdog.js` resident, 24 real KVS keys set):

| version | minified | result |
|---|---|---|
| `main` @ `ae4c5da` | 36155 | runs, `mem_peak 22330`, `mem_free 10276` |
| first fix | 40002 | `out_of_memory` |
| after static size trim | 38483 | `out_of_memory` |
| **after call-slot pool** | — | runs, **`mem_peak 21280`**, `mem_free 8148` |

`main` alone already peaks at 22330 of 23030 — ~700 bytes of headroom. So this was never a
code-size problem: trimming 1519 minified bytes changed nothing. It was **peak allocation** — the
wrapper allocated a fresh closure on every `Shelly.call` (39 during init alone). Replacing that with
a fixed pool of records claimed in place and passed via `userdata` to one shared completion handler
brought peak ~1050 bytes *below* `main`'s.

## Blocked on real hardware — do not merge yet

The final build runs on the Plus 1 but **still fails with `out_of_memory` on the production Pro1
(`filtration-hiver`)**. The devices report the same total heap but run **different firmware
generations — Pro1 on 2.0.0, Plus 1 on 1.7.5** — so the Plus 1 is not a valid proxy for this
particular constraint, despite matched scripts and matched KVS config.

Remaining options, in the maintainer's hands:
1. Free heap on the Pro1 (e.g. stop/remove `watchdog.js`, 3554 bytes) — a production behaviour change.
2. Reduce `main`'s own footprint so the whole thing fits with margin.
3. Split this PR: land Bug A (the crash fix, cheap) and defer Bug B (`restartCatchUp`, ~1.1 KB).

The production device is currently on v0.11.9's `pool-pump.js`, which is stable there
(`mem_peak 17136`, ~6 KB headroom) and completed a full unattended day cycle (10:18 start,
17:42 stop) during this work.

## Relationship to #401 / #406

#406 (removing the legacy `SolarAutomation` daemon code) is gated on #402/#403/#405 being verified
against the real pump. That gate is **not met** while this cannot run on the Pro1.

## Commits

- fix(pool-pump): bound in-flight RPCs and catch up on restart (#421) ([5ea7f73](https://github.com/asnowfix/home-automation/commit/5ea7f730a90a58797e22750e695939979f070a31))
- fix(pool-pump): fail loudly if the Shelly.call wrapper is not live (#421) ([75e161f](https://github.com/asnowfix/home-automation/commit/75e161f914b25a977ce3d760ba2af746a719bb4a))
- fix(pool-pump): shrink #421 fix's minified footprint for the Pro1 heap ([af8c9fc](https://github.com/asnowfix/home-automation/commit/af8c9fceb62ffffdc9eebee66d712146ee928ff1))
- fix(pool-pump): reuse call slots instead of a closure per RPC (#421) ([0ee8eac](https://github.com/asnowfix/home-automation/commit/0ee8eac4e980b82d98686bf587e96ed7cc36945d))

### DIFF
--- a/docs/421-pool-pump-crash-and-catchup-plan.md
+++ b/docs/421-pool-pump-crash-and-catchup-plan.md
@@ -1,0 +1,86 @@
+# Plan: issue #421 — pool-pump "Too many calls in progress" crash + restart catch-up
+
+Tracks implementation of https://github.com/asnowfix/home-automation/issues/421.
+Two independent bugs in `internal/shelly/scripts/pool-pump.js`, plus the minimal
+emulator capability (a slice of #250) needed to reproduce Bug A in
+`pkg/shelly/script/run.go`.
+
+Method: TDD / Prove-It. Emulator capability first, then failing tests proving
+both bugs reproduce against current `main`, only then the fixes.
+
+## Phase 1 — Emulator capability (prerequisite for reproducing Bug A)
+
+- [x] `pkg/shelly/script/device_state.go`: `ExecutionMode` type
+      (`DeviceTestMode` zero value / `DeviceExtensionMode`), `MaxConcurrentCalls`
+      const, `DeviceState.Mode` and `DeviceState.CallDelay` fields.
+- [x] `pkg/shelly/script/run.go`: per-VM `callsInFlight` counter around the
+      `Shelly.call` dispatch closure. `DeviceTestMode` panics with a message
+      matching the real device's `Too many calls in progress` once 5 calls are
+      already in flight. `CallDelay` (only meaningful in `DeviceTestMode`)
+      defers the *release* of a call's in-flight slot rather than the call's
+      dispatch/callback — this lets tests create realistic overlapping
+      in-flight calls without disturbing any script's existing synchronous
+      callback-ordering assumptions. Inert (zero overhead, zero behavior
+      change) in `DeviceExtensionMode` and whenever `CallDelay == 0`.
+- [x] `pkg/shelly/script/run.go`: `RunWithDeviceState`'s event loop now treats
+      *any* handler `Handle()` error as a full script crash (stops the loop,
+      returns the error) instead of logging and continuing. This is required
+      infrastructure, not scope creep: on real Shelly firmware an uncaught
+      exception in *any* callback (timer tick, event handler, MQTT message)
+      kills the whole script, and Bug A's real crash happens inside a
+      `Timer`-driven task-queue tick, not during the initial synchronous
+      script evaluation — the emulator can't observe "the script crashed"
+      for that path without this fix.
+- [x] Do NOT implement the rest of #250 (timer/event-handler/status-handler/
+      MQTT-subscription count ceilings) — only the concurrent-call slice
+      needed here. Cross-referenced in code comments.
+
+## Phase 2 — Bug A: reproduce, then fix
+
+- [x] `internal/shelly/scripts/pool_pump_test.go`: new test drives a normal
+      boot (reuses the existing `TestPoolPump_InitVerifiesSchedules` fixture
+      shape) with `Mode: DeviceTestMode` (default) and a `CallDelay` long
+      enough that config-loading's sequential `KVS.Get` calls (one dispatched
+      per 200ms task-queue tick, ~14 of them) overlap past the 5-concurrent
+      ceiling before the fix exists. Confirm it fails against current `main`
+      with a "Too many calls in progress" error — report the exact output.
+- [x] Fix (`pool-pump.js`): a single choke point (`Shelly.call` is reassigned
+      once, near `TASK_QUEUE`, to wrap every callback via `trackCall`) tracks
+      real script-wide `CALLS_IN_FLIGHT`. `processTaskQueue` throttles its own
+      dispatch to `MAX_CALLS_IN_FLIGHT` (4, one below the real ceiling),
+      retrying the same task on the next tick instead of assuming 200ms was
+      "long enough". This is the root-cause fix — no call site needed manual
+      edits.
+- [x] Second, independent layer of defense: `task()` in `processTaskQueue` is
+      wrapped in `try/catch` — a queued task throwing no longer kills the
+      whole script.
+- [x] Re-run the new test — now passes (init completes, no crash).
+
+## Phase 3 — Bug B: reproduce, then fix
+
+- [x] New test: boot with a schedule window (`Schedule.List` fixture) that
+      excludes "now" and the switch already "on" in `ComponentStatus`. Assert
+      the fixed `enforceOutputState()` path calls `doStop()` and
+      `active-output` ends up `-1` without waiting for any schedule tick
+      (there is no cron emulation in this harness).
+- [x] Fix (`pool-pump.js`): new `restartCatchUp()`, queued right after the
+      existing init steps complete (after `verifySchedules`, so it reads a
+      validated `Schedule.List`), compares current time-of-day against the
+      currently active mode's window (mode-aware: summer →
+      handleMorningStart/handleEveningStop, winter →
+      handleNightStart/handleNightStop) and calls `doStart()`/`doStop()` if
+      the physical state disagrees. Skips gracefully (no action) when the
+      window can't be resolved (e.g. still-symbolic `@sunrise` timespecs on a
+      schedule that has never had `updateScheduleMode()` run against it).
+- [x] Regression guard: three existing tests intentionally leave the pump
+      "on" in `ComponentStatus` at boot to test *other* behavior (water-supply
+      protection, runtime-accounting continuity). Bug B's fix would otherwise
+      immediately stop them, since their shared schedule fixture's night
+      window (23:15–00:15) essentially never contains real test-run wall-clock
+      time. Added `activeNightWindowSchedules()` (widens that window to
+      bracket `time.Now()`) and pointed those three tests at it.
+
+## Phase 4 — Full regression pass
+
+- [x] Re-run all of `pool_pump_test.go`.
+- [x] `make test` (canonical, root + all workspace sub-modules).
--- a/internal/shelly/scripts/pool-pump.js
+++ b/internal/shelly/scripts/pool-pump.js
@@ -438,10 +438,12 @@ var STATE_KEYS = {
 // arguments it always did (trackCall forwards result/error_code/
 // error_message/userdata unchanged), so this is purely additive bookkeeping.
 var CALLS_IN_FLIGHT = 0;
+var CALLS_TRACKED = 0;       // lifetime count — proves tracking actually runs
 var MAX_CALLS_IN_FLIGHT = 4; // stay one below the real 5-call device ceiling
 
 function trackCall(onDone) {
   CALLS_IN_FLIGHT++;
+  CALLS_TRACKED++;
   return function (result, error_code, error_message, userdata) {
     // Decrement unconditionally before invoking onDone, so a slot is always
     // freed even if onDone itself throws (caught by processTaskQueue's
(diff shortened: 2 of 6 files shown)


asnowfix:
**2026-08-10 20:47 CEST** — Closing: everything this PR set out to deliver is now on `main`, and its remaining diff is actively dangerous.

**What it carried, and where that landed instead:**

| This PR | Now on main via |
|---|---|
| Restart catch-up / schedule reconciliation | #442 (`reconcileRunState`, `isWithinRunWindow`) |
| Heap reduction to fit the device | #447 (config schema SoT, −3584 B measured) and #458 (single controller, −2240 B) |
| In-flight RPC bounding (#421) | **#461**, ported from the form running in production |

**Why not rebase it.** The remaining delta is the **pre-#450 `sharedCallDone`**, which invokes the caller's callback synchronously inside the native completion. That is the mechanism behind the crash that killed the production pump three times in 26 hours with `Too much recursion — the stack is about to overflow`. #461 brought the same machinery to `main` *with* the `MAX_CALL_DEPTH` guard; merging this PR would put the defective version back.

It is also `DIRTY` against ten merges from 2026-08-09/10.

**Note for anyone reading this later:** the depth guard in #460/#461 did **not** cure #450 — v0.12.0 crashed again on 2026-08-10 at 16:44 with the guard demonstrably live on the device. #450 remains open, and the pump is now instrumented (UDP debug capture plus a 2-minute watcher) to catch the next occurrence with the script's own log lines rather than four truncated stack frames.

Issue #422, opened 2026-08-03
test: establish real-device integration test campaign using spare Shelly Plus 1 (development)


asnowfix:
## Background

This repo currently has two tiers of pool-pump script testing, with a gap between them:

1. **Emulator-based unit tests** (`internal/shelly/scripts/pool_pump_test.go`, running against
   `pkg/shelly/script/run.go`'s goja-based Shelly script emulator) — fast, deterministic, no real
   hardware, but the emulator doesn't model every real-device constraint (see #250, open: it
   currently enforces no resource limits — timers, event/status handlers, MQTT subscriptions,
   concurrent RPC calls — at all).
2. **Ad-hoc manual live-device verification** — a human or coding agent SSH/MQTT-ing into a real
   device during a one-off session, with no repeatable record of what was checked or how (see
   `docs/401-campaign-live-verification-status.md` for an example write-up of one such session,
   including a crash discovered live and fixed manually on the spot).

There is nothing in between: no repeatable, low-risk way to run a defined integration test case
against **real Shelly hardware** without either (a) risking the actual production pool pump device
(`filtration-hiver`), or (b) re-deriving the whole test procedure from scratch each time in a fresh
ad-hoc session.

A spare Shelly Plus 1, device name `development`, is available to serve as a dedicated integration
test target — safe to crash, reboot, or reflash repeatedly without affecting any real home
automation (pool pump, garden sprinklers, etc.). Device-type auto-detection in `pool-pump.js`
(`internal/shelly/scripts/pool-pump.js:321-331`) is purely by switch count — a single-switch Plus 1
takes the exact same code path as the Pro1 used by `filtration-hiver`, so pump scripts under test
need **no code changes** to run on it; only device-side config differs (KVS `preferred` device ID,
test-specific `Schedule` jobs instead of daily-cadence ones).

## Goal

Establish a lightweight, repeatable **integration test campaign** — a structured place to define,
record, and re-run test cases against real hardware using `development` as the guinea pig — without
building a full hardware-in-the-loop CI framework (out of scope; this is a hobby project, keep it
simple). Concretely:

1. Add `docs/integration-tests.md` (new file) as the campaign's living record. Each test case gets
   a section using this template:
   - **Script/version under test** (git ref or PR #)
   - **Purpose** (what real-hardware behavior this verifies that the emulator can't)
   - **Preconditions** (required KVS config, `Schedule` jobs, firmware version)
   - **Steps** (exact commands / RPC calls to run)
[...]

PR #427, opened 2026-08-04
test(422): real-device integration test campaign + #421 baseline

asnowfix:
Addresses #422.

Adds `docs/integration-tests.md` — a living record for real-hardware test cases run against the
spare Shelly Plus 1 (`development`, `shellyplus1-08b61fd98f44`, fw 1.7.5), using the template
specified in #422 (script/version under test, purpose, preconditions, steps, pass/fail criteria,
originating issue, last run).

Test case #1 is #421's real-hardware verification, filled in with the commands actually run and the
results actually observed — including the live reproduction of the `Too many calls in progress`
crash at a **third** call site beyond the two #421 documented (`storeValue("runtime-sec", ...)` in
`persistRuntimeState`).

## Findings from establishing the rig, recorded in the doc

- **The JS heap is one shared pool across all scripts on a device (~23 KB).** `pool-pump.js` will not
  even start unless other resident scripts are stopped to make room. Confirmed empirically:
  stopping `watchdog.js` and `myhome-link.js` moved `mem_free` from 21350 to 25200.
- **A spare device is only a valid proxy if its configuration matches production.** With 5
  `script/pool-pump/*` KVS keys the script used `mem_used 13748`; with production's 24 keys the same
  code OOM'd. Roughly 7.9 KB of footprint is config-driven runtime state. Any memory testing done
  without replicating production KVS is falsely easy.
- **Firmware generation matters and is not interchangeable.** The production Pro1 runs fw 2.0.0 and
  the spare Plus 1 runs fw 1.7.5; a build that runs on the Plus 1 still OOM'd on the Pro1 despite
  matched scripts and matched KVS.
- **`nc -u -l -k` is unreliable for UDP debug capture on macOS 15.7.7** even with `-k` — it silently
  stops recording after the first burst. Use a small Python UDP listener instead, and always prove
  it with test datagrams before trusting a long capture. Device debug logs are NUL-separated
  (`tr '\0' '\n'`).
- **The device truncates each UDP debug line at ~128 characters**, so on-device diagnostics must be
  written as short lines or the actionable part is silently cut off.
- **`ctl shelly script upload` takes a bare script name from the binary's embedded FS**, not a
  filesystem path — so uploading a specific version means building a binary from that version.

The doc includes an explicit **Restore steps** section (restart the stopped scripts, remove the test
Schedule jobs, clear the test KVS keys) so a cold reader can undo the rig.

## Note

Test case #1's outcome is recorded as the baseline reproduction plus the fix verification performed
on the Plus 1. The fix itself is **not** yet verified on the production Pro1 — see #426, which is
blocked on device memory. This PR only records the campaign and its process; it does not claim #421
is done.

## Commits

- test(422): establish live-hardware rig on development, record #421 baseline ([7506965](https://github.com/asnowfix/home-automation/commit/7506965af5cede46da4fd19ce6833e6393b6b626))
- test(422): reproduce #421 crash on development after freeing shared JS heap ([ddd6acf](https://github.com/asnowfix/home-automation/commit/ddd6acf0bb8a8d3ee2cfbcc4aaad9af854005a8a))

### DIFF
--- a/docs/integration-tests.md
+++ b/docs/integration-tests.md
@@ -33,9 +33,25 @@ is a point-in-time snapshot, not live truth.
   Never confuse with `filtration-hiver` (`shellypro1-ec62608c0230`, the real pool pump — a Shelly
   **Pro1**, different hardware family, more script RAM — see the "Plus 1 vs Pro1/Pro3 memory" note
   in test case #1 below).
-- It already runs two permanent scripts that must never be stopped or deleted by a test case:
-  script id 1 `watchdog.js`, script id 2 `myhome-link.js`. Test scripts under evaluation should
-  land at script id 3+ (whatever the device assigns next).
+- It already runs two permanent scripts, script id 1 `watchdog.js` and script id 2
+  `myhome-link.js`. Test scripts under evaluation should land at script id 3+ (whatever the device
+  assigns next). **Deleting** either permanent script is still off-limits for any test case (it's
+  real config the device is meant to run day-to-day) — but **stopping** them temporarily turned out
+  to be *required*, not just tolerated, for test case #1 below: see "Shared JS heap" immediately
+  below. Always restart both (`ctl shelly script start development watchdog.js` /
+  `myhome-link.js`) at the end of a session that stopped them — a test case is not "done" while
+  they're down.
+- **Shared JS heap is small and shared across every script on the device**: this Plus 1's Gen2 JS
+  engine gives all scripts on the device **one shared heap pool** (not one pool per script) —
+  confirmed empirically in test case #1 by watching `Script.GetStatus`'s `mem_free` move in lockstep
+  across script ids 1, 2, and 3 as scripts were stopped/started. Total pool size is roughly 25 KB
+  (`watchdog.js` ~2.1 KB + `myhome-link.js` ~1.7 KB + ~21.3 KB free = ~25.2 KB with both idle-running;
+  a script's own `mem_peak` while compiling/running eats further into whatever's left). A `pool-pump.js`
+  build whose minified size approaches or exceeds the free remainder at upload time will fail to
+  even compile (`out_of_memory`, zero log output, before the script's first line runs) if the other
+  two scripts are left running. **Any future test case that uploads a script close to or above
+  ~20 KB minified onto `development` must first stop `watchdog.js`/`myhome-link.js` to free enough
+  heap**, and must restart them afterward.
 
 Environment gotchas that affect every test case run from a macOS dev machine on this network:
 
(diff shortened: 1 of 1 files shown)


asnowfix:
## Rebased onto current `main` and substantially extended — 2026-09-06

Untouched since 2026-08-20 and **210 commits behind**, though it applied with **zero conflicts** (the
file does not exist on `main`). Original two commits preserved as cherry-picks; refresh on top.
Documentation only. `docs/integration-tests.md` goes from 463 to **772 lines**.

**Why this is worth landing rather than closing.** The hardest recurring question in this repo is *"was
this actually verified on hardware, and on which hardware?"* — and nothing in the tree answers it. That
matters more now than when the PR was opened, because the fleet's two devices are **not**
interchangeable: `filtration-hiver` is a Pro1 on fw **2.0.0** and is the production pump, while
`development` is a Plus 1 on fw **1.7.5** — a different app *and* firmware major. **There is no
Pro1-on-2.0.0 test bed**; the only instance of that combination is production. Both re-verified by
read-only `curl` today.

### Four campaign runs added, each sourced to a dated comment, device and firmware

| case | what | device | when |
|---|---|---|---|
| #2 | Gate 3 — solar extension and stop | `filtration-hiver` | 2026-08-24 |
| #3 | Gate 3 — solar start | `filtration-hiver` | 2026-08-30 12:53:54 |
| #4 | `ReferenceError` inside `try` — containment **corrected** on both platforms | `development` *and* `filtration-hiver` | 2026-09-02 |
| #5 | KVS key length limit, 42 chars, error `-103` | `mezzanine` | 2026-08-23 |

Case #4 is the one that best justifies the whole document: it records a previously-stated "hard rule"
that **did not reproduce** when actually measured on both platforms. That correction existed only in
issue comments until now.

### A new section: "What is NOT verified on hardware — first-class, not a footnote"

Its first entry is **#587 / commit `f380a143`, merged yesterday and never run on any device.** It
records the pre-deployment baseline (`mem_free` 7280, `mem_peak` 22918, 2026-09-05 09:58 CEST) and the
+636 minified-byte cost — and then, correctly, refuses to let that stand in for a heap figure:

> **Do not infer this change's resident cost from +636 bytes.** Source bytes and resident heap cost do
> not correlate linearly — deletions have measured as low as **15%** of their byte delta in resident
> cost, and single-call-function inlining as high as **167%**.

That is a sharper caution than I had been applying myself, and it is the right one.

### Also recorded

- The **upload-path blocker**: `ctl shelly script upload` and the `shelly` MCP server both resolve
  devices through a daemon on instance `local`, which is not running (`device.lookup` timeout). Starting
  one is not a free workaround — it also publishes `myhome/energy/solar/available`, a second voice into
  the live pump's solar input. This is what currently prevents *any* device test from running.
- A dated correction rather than a silent rewrite: `development`'s "left crashed, not running" end-state
  from the August content is stale; a later #401 session left `pool-pump.js` running and
  `myhome-link.js` stopped with `out_of_memory`. The historical record is preserved with a
  2026-09-06 note beside it.

### Verification and honest gaps

`make generate && make build` clean. **`make test` deliberately not run** — docs-only, and the suite's
slowest package costs ~10 minutes.

Cases #2–#5 were **mined from issue history, not re-run** this session; the only device interaction was
the read-only `curl` checks. Every figure is either dated-and-sourced to a specific comment or
explicitly marked as not yet measured. One candidate case (#589's `ctl pool add`/`pool update` mutation
of live config, 2026-09-02) was **left out rather than rushed** at the end of the time budget — worth
adding in a follow-up.

`#422` stays **open**: it asks for an ongoing campaign, which a single merged document does not
discharge.

asnowfix:
## ⚠️ Correction, 2026-09-06 — "there is no working upload path" was wrong

I asserted earlier today, here and in several other places, that device work was blocked because
`ctl shelly script upload` and the `shelly` MCP server resolve devices through a daemon on instance
`local` that is not running. **The premise was wrong, and it was wrong in the worst direction: it
deterred hardware testing that was available all along.**

`LookupDevices()` in `internal/myhome/client.go` short-circuits **before** contacting any daemon. A
`.local` suffix is mDNS-resolved to an address; an address goes straight to
`shelly.NewDeviceFromIp()`. Only a **bare name** falls through to `hc.start(ctx)` and the
`device.lookup` RPC.

Verified 2026-09-06, no daemon running:

```
$ ./myhome-local.sh shelly script list filtration-hiver.local    # works
- id: 1  name: watchdog.js   running: true
- id: 2  name: pool-pump.js  running: true

$ ./myhome-local.sh shelly script list 192.168.1.21              # works, same output

$ ./myhome-local.sh shelly script list filtration-hiver          # FAILS
timeout waiting for response to method device.lookup after 14s (dst: local)
```

**So the correct instruction is: address the device by `.local` name or IP.** No daemon needs to be
started, and in particular nobody should start one as a workaround — it would publish
`myhome/energy/solar/available`, a second voice into the live pool pump's solar input.

mDNS is not uniformly reliable here, so keep the IP as a fallback: `development.local` resolved via
`ping` but not `curl` earlier today, needing `192.168.1.30` directly. `filtration-hiver` is
`192.168.1.21`.

The ergonomics problem that remains — a bare name failing with a 14-second timeout and an error
message that names neither cause nor remedy — is filed as its own issue. The blocker, as described, was
not real.

asnowfix (commented):
LGTM

Issue #367, opened 2026-07-14
logging & deps: replace hlog global/init variants (evaluate log/slog), slim Google Cloud SDK and YAML dependencies


asnowfix:
## Context

From a full Go expert review (2026-07-13, main @ 2cd3628). Logging and the dependency tree carry more weight than the project needs.

## Logging (`hlog`)

- `hlog.Logger` is a mutable package-level global set by four near-identical exported Init variants (`Init`, `InitWithDebug`, `InitForDaemon`, `InitForDaemonWithDebug` — all one-liners into `InitWithLevel`). Any package calling `hlog.GetLogger()` at init/package-var time gets a zero `logr.Logger` — an init-order trap.
- The stack is logr → zerologr → zerolog + lumberjack + kardianos/service (the latter used inside hlog only to answer "am I interactive?"). Since Go 1.21, `log/slog` covers structured logging in the stdlib; logr has an official slog bridge if the logr API should stay during migration.
- Environment sniffing surprises: `VSCODE_PID`/`VSCODE_IPC_HOOK` silently force debug level (`hlog/hlog.go:117-125`) — running from a VS Code terminal changes behavior. Make it an explicit flag/env var.
- File-rotation via lumberjack may be unnecessary: under systemd, journald already captures stderr (the code detects this, `hlog/hlog.go:175-178`); macOS/Windows services similar. Decide whether file logging is needed at all outside "no service manager" cases.

Proposed: single constructor `hlog.New(opts) (logger, error)` (no global, no panic — see also the error-handling conventions issue), injected from the composition root (see DI issue); then evaluate replacing the zerolog stack with `log/slog` + `slog-logr` bridge, or fully migrating call sites to slog.

## Dependency slimming

- **Google Cloud SDK for one feature:** `internal/myzone/gcp_dns.go` pulls `google.golang.org/api` → gRPC + OpenTelemetry + protobuf + cloud.google.com/auth into the root binary — megabytes of binary and the largest supply-chain surface in the project, for Cloud DNS record updates. The Cloud DNS v1 REST API is a couple of `net/http` calls with an OAuth2 token (`golang.org/x/oauth2/google` alone suffices). Evaluate replacing the SDK.
- **Three YAML libraries** in the graph: `gopkg.in/yaml.v2`, `go.yaml.in/yaml/v3`, `sigs.k8s.io/yaml`. After the module collapse, converge deliberately on one.
[...]

Issue #428, opened 2026-08-04
bug(script upload): successful uploads reported as failures when post-upload RPC times out


asnowfix:
## Summary

`myhome ctl shelly script upload` reports **failure** for uploads that actually **succeed**. The
chunked `Script.PutCode` transfer completes and the correct code lands on the device, but a
follow-up RPC in the same command (`KVS.Get` for version tracking, and/or `Script.Start`) times out,
so the command exits non-zero with:

```
✗ Failed to upload pool-pump.js to filtration-hiver: timeout waiting for response from shellypro1-ec62608c0230 (filtration-hiver)
✗ 1 device(s) failed:
Error: all 1 device(s) failed
```

The script is nonetheless present and correct on the device, and often already running.

## Why this matters

This is not cosmetic. During live debugging of #421 on 2026-08-04 this misleading failure caused a
**wrong conclusion about real hardware**: an upload was reported as failed, the script was then
found not running, and the result was recorded as "this build cannot run on the Pro1 — it OOMs."
A later retry of the *same build* ran fine (`mem_peak 21350`, `mem_free 6398`). The false failure
report made a marginal-but-working configuration look like a hard hardware limit, and sent the
investigation down a multi-hour detour into memory optimisation that may not have been needed.

## Reproduction

Device: `filtration-hiver`, id `shellypro1-ec62608c0230`, Shelly **Pro1**, firmware **2.0.0**,
reached over MQTT (`--mqtt-broker tcp://192.168.1.2:1883`). Script: `pool-pump.js` (~37-40 KB
minified), uploading over an existing script id 2 with `--force`.

```bash
go run ./myhome ctl shelly script upload --mqtt-broker tcp://192.168.1.2:1883 -T 90s --force filtration-hiver pool-pump.js
```

Observed on **every one of ~5 attempts** on 2026-08-04, with `-T 60s`, `-T 90s` and `-T 120s`.
Verify the upload actually succeeded with:

```bash
go run ./myhome ctl shelly script call --mqtt-broker tcp://192.168.1.2:1883 -T 90s filtration-hiver Script.GetStatus '{"id":2}'
```

The reported `mem_used`/`mem_peak`/`mem_free` identify which version is loaded, and
`Script.GetCode` can be grepped for version-specific identifiers to confirm.

Note the smaller/older v0.11.9 build of the same script (~30 KB minified) also triggers it, so this
is not purely a function of script size.

## Likely cause

[...]

PR #430, opened 2026-08-05
feat(minifier): esbuild engine with opt-in top-level mangling and symbol map

asnowfix:
Adds `github.com/evanw/esbuild/pkg/api` as a second, **selectable** JS minifier alongside tdewolff,
with optional top-level identifier mangling and a symbol map for decoding crash traces.

**Draft on purpose**: nothing is enabled by default, and the mangling path has not been verified on
real hardware. This repo auto-merges PRs when CI goes green, so it stays draft until someone
deliberately device-tests it.

## Why

The Shelly JS heap is one pool shared across all scripts on a device (~23 KB). `pool-pump.js` on the
production Pro1 runs at `mem_peak 21350` — roughly 1680 bytes of headroom — and has failed to start
with `out_of_memory` more than once (#421 / #426). tdewolff mangles local identifiers only; top-level
names survive untouched, and on `pool-pump.js` they account for ~121 symbols over ~899 occurrences.

## Measured results

| script | raw | tdewolff | esbuild (no mangle) | esbuild + top-level mangle |
|---|---|---|---|---|
| pool-pump.js (main) | 76614 | 36155 | 36162 | **31010** |
| pool-pump.js (#426 branch) | 90065 | 37739 | 37601 | **31318** |
| watchdog.js | 8186 | 3338 | 3315 | **3193** |
| garden.js | 44363 | 23971 | 23743 | **19493** |
| prometheus-metrics.js | 11338 | 6075 | 6063 | 6032 |

For reference v0.11.9's `pool-pump.js` — which runs comfortably on the Pro1 at `mem_peak 17136` — is
30409 bytes. Mangling brings the current script back into that range (−14%), and the heavier #426
version from an OOM-crashing 37739 down to 31318 (−17%).

## How top-level mangling works

esbuild will not rename top-level names in a plain script — they are globals. The script is wrapped
in an IIFE whose return value re-exports the preserved names at true top level:

```js
var __myhome_minify_exports__=(function(){ /* script */ return {handleMorningStart:tt,...}; })();
handleMorningStart=__myhome_minify_exports__.handleMorningStart, ...
```

This is required for correctness: device `Schedule` jobs invoke `handleMorningStart()`,
`handleEveningStop()`, `handleNightStart()`, `handleNightStop()` and `handleDailyCheck()` **by name**
via `script.eval`. Mangling them would silently stop every schedule firing. The preserve-list is
per-script and configurable (`internal/shelly/scripts/minify_config.go`), not hardcoded.

## Symbol map and `demangle`

Shelly crash traces carry function names and code snippets but **no line/column**, e.g.
`in function "storeValue" called from storeValue("turnover-today",...)`. A Source Map v3 alone
cannot resolve that. So the Source Map's `names` array is post-processed into a flat
mangled -> original table for top-level symbols (unambiguous — they share one flat scope), and
`myhome ctl shelly script demangle` rewrites a crash message with it.

Verified end-to-end: `in function "P"` -> `in function "storeValue"`.

## Safety

- **Default OFF everywhere.** `scripts.minifier_engine` defaults to `tdewolff` and
  `scripts.mangle_top_level` to `false`, reproducing today's behaviour exactly. This matters because
  the minifier runs *inside the daemon* — `manager.go` -> `shellysetup.SetupDevice()` uploads
  `watchdog.js` automatically to every discovered device, with no human in the loop.
- `Target: api.ES5` is mandatory; esbuild would otherwise emit ES6 into an Espruino engine.
- `ComputeScriptVersion` hashes the raw embedded source, so changing engines cannot churn version
  hashes or trigger mass re-uploads. Verified.
- esbuild is cgo-free: `GOOS=linux GOARCH=arm64 go build` produces a statically linked aarch64
  binary. Cost: **+4.86 MB** (38.97 MB -> 43.83 MB), relevant to #367.

## Known limitations

- **9 of 19 embedded scripts use `let`/`const`** and esbuild's ES5 target cannot downlevel them — it
  errors loudly rather than emitting wrong code (safe failure, but those scripts cannot use this
  path today).
- With mangling on, any top-level name not preserved or debug-exported becomes unreachable via live
  `Script.Eval` — the exact capability the #421 investigation relied on. `DebugExports` mitigates
  this but is opt-in per name.
- The daemon's automatic upload path does **not** persist a symbol map; enabling mangling for real
  uploads means separately running `script minify --symbol-map` to retain crash decoding.
- `Demangle` is word-boundary textual substitution, not a parser.

## Bonus finding on #266

esbuild's default `MinifySyntax` **reproduces** tdewolff's crash-inducing comma-expression-in-return
pattern — it does not sidestep #266. A new `DisableSyntaxLowering` option keeps statements
semicolon-separated and structurally avoids the pattern for ~180 bytes. **Static analysis only, not
hardware-verified.** If it holds up on a device, PR #267 would likely be superseded or much
simplified.

## Test plan

- [x] `make generate && make build`
- [x] `make test` — 47 packages ok, 0 failures (reviewer-verified independently)
- [x] `GOOS=linux GOARCH=arm64 go build -o /tmp/x ./myhome` — statically linked aarch64
- [x] Preserved globals survive mangling (all five present in output)
- [x] Symbol map round-trip on a real #421 crash string
- [ ] **Device-test at least `pool-pump.js` and `watchdog.js` before enabling in production**

## Commits

- feat(minifier): add esbuild engine with opt-in top-level mangling ([945a42f](https://github.com/asnowfix/home-automation/commit/945a42f159407d9f8253df48989f4f1031b79a70))
- fix(minifier): emit raw UTF-8 from esbuild, not \uXXXX escapes ([5f9afea](https://github.com/asnowfix/home-automation/commit/5f9afea421d7063cc825f2727c7f669171405dcf))
- fix(minifier): decode source-map columns as UTF-16 units, not bytes ([f0c0118](https://github.com/asnowfix/home-automation/commit/f0c0118b02296bea6e6a7d075efac477ebfdd15f))

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -621,6 +621,71 @@ pool:
 | `pool.solar.min_volume_turnover` | `MYHOME_POOL_SOLAR_MIN_VOLUME_TURNOVER` | `--pool-solar-min-volume-turnover` | `5` | Soft-stop target: pool volumes filtered per day; pump keeps running past this while solar is still above the start threshold |
 | `pool.solar.max_volume_turnover` | `MYHOME_POOL_SOLAR_MAX_VOLUME_TURNOVER` | `--pool-solar-max-volume-turnover` | `7` | Hard ceiling: pool volumes filtered per day; pump always stops (and won't be solar-started) once reached |
 
+## Scripts (JS minifier)
+
+The daemon's automatic device setup path (triggered whenever it discovers an
+unconfigured Shelly device on the network) uploads `watchdog.js` with
+minification enabled and no human in the loop. These options select which
+minifier it uses. **Both default to today's exact, long-proven behavior** —
+`tdewolff`, no top-level identifier mangling — and must be changed
+explicitly; see the "Default OFF" rationale below.
+
+### Example
+
+```yaml
+scripts:
+  minifier_engine: "tdewolff"   # or "esbuild"
+  mangle_top_level: false        # esbuild only
+```
+
+### Options
+
+| Key | Env var | Flag | Default | Description |
+|-----|---------|------|---------|-------------|
+| `scripts.minifier_engine` | `MYHOME_SCRIPTS_MINIFIER_ENGINE` | `--script-minifier-engine` | `tdewolff` | JS minifier for the daemon's automatic watchdog.js upload: `tdewolff` (proven-safe, mangles local identifiers only) or `esbuild` (experimental; ES5-targeted, can also mangle top-level identifiers) |
+| `scripts.mangle_top_level` | `MYHOME_SCRIPTS_MANGLE_TOP_LEVEL` | `--script-mangle-top-level` | `false` | Also mangle top-level (module-scope) identifiers; only takes effect with `minifier_engine: esbuild`. Reduces output size further (measured ~19% smaller on pool-pump.js) but wraps the script in an IIFE, which makes internal top-level names (e.g. `STATE`, `CONFIG`) unreachable from a live `Script.Eval` session unless separately declared as debug-exports in code (see `internal/shelly/scripts/minify_config.go`) |
+
+### Why default OFF
+
+Both the `esbuild` engine and `mangle_top_level` are new (see the minifier
+work tracked from issue-derived exploration around pool-pump.js's JS-heap
+pressure). The daemon's auto-setup path
+(`myhome/devices/impl/manager.go` → `internal/myhome/shelly/setup.SetupDevice`)
+runs unattended the moment a new, unconfigured device shows up on the
+network — a minifier bug there would silently push a broken script to real
+hardware. Both options must be flipped on deliberately, by a human editing
+config or passing a flag, never by a version upgrade alone. `myhome ctl
+shelly setup` also accepts the same two flags for one-off manual setups.
+
+A related, per-script safety mechanism lives in code, not config: which
+top-level names must survive `mangle_top_level` as real globals (because a
+Shelly `Schedule` job calls them by name via `script.eval`) is a per-script
+list in `internal/shelly/scripts/minify_config.go`
+(`scriptMinifyOverrides`), not a global setting — get it wrong for a given
+script and that script's scheduled automation silently stops firing.
+
+### CLI: minify and demangle
+
+```bash
+# Minify an embedded script locally (no device involved) and report size:
+myhome ctl shelly script minify pool-pump.js --engine esbuild --mangle-top-level
+
+# Same, also writing the minified code and a symbol map:
+myhome ctl shelly script minify pool-pump.js --engine esbuild --mangle-top-level \
+  --out /tmp/pool-pump.min.js --symbol-map /tmp/pool-pump.symbols.json
+
+# Translate a Shelly crash message's mangled names back to source names:
+myhome ctl shelly script demangle /tmp/pool-pump.symbols.json \
+  'in function "t" called from t(n.count)'
+```
+
+The symbol map is a flat `mangled -> original` table for top-level symbols
+only (see the doc comment on `pkg/shelly/script.SymbolMap`) — it exists
+because Shelly crash traces give function names and code snippets but never
+a line/column, so an ordinary Source Map v3 (also generated, alongside the
+symbol map, when `--mangle-top-level` is set) cannot resolve them on its
+own.
+
 ## Notice & Email (SMTP)
 
 The notice service (see `docs/notice-events-plan.md`) curates a `notice` severity for events worth a human's attention — the daily pool/garden plans, solar pump on/off, and motion at night or while the home is unoccupied — and emails a daily digest. Unlike the events/occupancy/temperature services, it is **not** auto-enabled with the device manager: it depends on both the events and occupancy services already running, so an operator opts in explicitly with `notice.enabled: true` or `--enable-notice-service`.
--- a/internal/myhome/shelly/script/ops.go
+++ b/internal/myhome/shelly/script/ops.go
@@ -36,6 +36,34 @@ func UploadNamedScript(ctx context.Context, log logr.Logger, via types.Channel, 
 // UploadWithVersion uploads a script and tracks its version in KVS
 // This is MyHome-specific business logic that combines script upload with version tracking
 func UploadWithVersion(ctx context.Context, log logr.Logger, via types.Channel, device types.Device, name string, code []byte, minify bool, force bool) (uint32, error) {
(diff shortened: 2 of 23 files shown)


asnowfix:
## Device verification on real hardware — one blocking bug found and fixed

Tested on the spare Shelly Plus 1 (`development`, fw 1.7.5) with `watchdog.js` resident and the 24 real production KVS keys replicated, i.e. production-equivalent load.

### Blocking bug: esbuild emitted `\uXXXX` escapes, which Espruino rejects

The mangled script would not parse on device at all:

```
Uncaught SyntaxError: \uXXXX literals are disallowed
 at if(t>=e.length){n("✓ All initialization steps complete ...
```

esbuild defaults to ASCII output and escapes non-ASCII as `\uXXXX`. `pool-pump.js` has a `✓` (U+2713) and five em dashes (U+2014) in log strings.

**Nothing in CI could have caught this.** goja accepts `\uXXXX`, so the suite stayed green — and the output actually looked *safer* by the pre-flight heuristic, because escaping had removed every non-ASCII byte and thereby eliminated the #423 chunk-splitting risk. The output was pure ASCII precisely because it was broken.

This affects the engine **with and without** mangling, so adopting esbuild at all — not just the mangling option — would have shipped unparseable scripts to every device through the daemon's automatic `watchdog.js` upload.

Fixed in 5f9afea: `Charset: api.CharsetUTF8` on both `api.Transform` call sites, plus `TestMinifyWithOptions_EsbuildEmitsRawUTF8NotEscapes` covering both paths (verified failing without the fix).

### Post-fix results

| | unmangled `main` | esbuild + mangled |
|---|---|---|
| minified | 36155 | **31010** (−5145) |
| `mem_peak` | 22330 | **21686** (−644) |
| `mem_free` | 10276 | **12040** (+1764) |

Functionally verified: `__myhome_minify_exports__` present (genuinely the mangled build); all five schedule-invoked globals report `function`; and `handleEveningStop()` was invoked **by name** exactly as a device Schedule job does, with the script surviving.

### Important caveat — this does not solve #433

**Top-level mangling buys ~644 bytes of peak, not ~5 KB.** Static size drops 5145 bytes but `mem_peak` barely moves, because peak is dominated by transient allocation rather than code size — the same lesson #421 taught (there, trimming 1519 static bytes changed nothing while removing one per-call closure moved peak ~1050 bytes).

So this PR should **not** be treated as the enabler that makes the solar path fit. #433 needs allocation-side work: dropping the unused `sources` field from the solar payload so the device parses less, and reducing the solar path's retained state.

### Incidental observations

- The upload reported `out_of_memory` but had **actually succeeded** — another instance of #428. Only dumping `Script.GetCode` revealed the truth.
- After repeated failed loads the device needed a reboot to clear fragmented heap before it would accept a new script.

Still draft: `watchdog.js` has not been device-tested under the esbuild path, and that is the script the daemon auto-uploads fleet-wide.

asnowfix:
## Triage 2026-09-06 — recommend **park, do not close**. Blocked on a gate, not on effort.

Reviewed while clearing the `stale-unreviewed` backlog. Untouched since 2026-08-20, **209 commits
behind**, but only **2 conflict regions** — cheap to keep alive.

**Still valuable, arguably more than when opened.** Bytes remain the binding constraint on this fleet.
`filtration-hiver` today: `mem_free` **7280**, `mem_peak` **22918** (read 2026-09-05 09:58 CEST). #587
had to be costed in minified bytes (+636) as recently as 2026-09-05, and every future pump change will
face the same arithmetic. tdewolff mangles locals only; this PR's ~121 top-level symbols over ~899
occurrences on `pool-pump.js` are exactly the headroom that is currently unavailable.

**Why it cannot simply be merged.** Its own description is right to hold it draft: the mangling path has
**not been verified on real hardware**, and this repo auto-merges when CI goes green. The verification
it needs is on a Pro1 running fw 2.0.0 — and **there is no such test bed**; `development` is a Plus 1 on
1.7.5, a different app *and* firmware major, so a green run there would not be evidence. The only
Pro1/2.0.0 in existence here is the production pool pump.

So this is gated on the same thing as #587's deployment: a deliberate, scheduled hardware session on
`filtration-hiver`, plus the upload path being usable at all (see below).

**One number in the description is now stale**: it cites `mem_peak 21350` with ~1680 bytes of headroom.
That was before #439/PR #447 dropped `CONFIG_SCHEMA` descriptions (measured `mem_peak` 22778 → 19194,
−3584 B) and before PR #526 removed unreachable schedule functions (−2530 B minified). Re-measure before
using the old figures to justify the change — the case is still good, but on today's numbers.

**Also note a current blocker to any device test**: `ctl shelly script upload` and the `shelly` MCP
server both resolve devices through a daemon on instance `local`, and none is running — calls fail with
`timeout waiting for response to method device.lookup after 14s (dst: local)`. Starting one is not free:
it also publishes `myhome/energy/solar/available`, a second voice into the live pump's solar input.

**Recommendation:** keep open and draft, rebase when someone schedules the hardware session. Do not
close — the analysis and the measured comparison table are worth more than the rebase will cost.

asnowfix:
## ⚠️ Correction, 2026-09-06 — "there is no working upload path" was wrong

I asserted earlier today, here and in several other places, that device work was blocked because
`ctl shelly script upload` and the `shelly` MCP server resolve devices through a daemon on instance
`local` that is not running. **The premise was wrong, and it was wrong in the worst direction: it
deterred hardware testing that was available all along.**

`LookupDevices()` in `internal/myhome/client.go` short-circuits **before** contacting any daemon. A
`.local` suffix is mDNS-resolved to an address; an address goes straight to
`shelly.NewDeviceFromIp()`. Only a **bare name** falls through to `hc.start(ctx)` and the
`device.lookup` RPC.

Verified 2026-09-06, no daemon running:

```
$ ./myhome-local.sh shelly script list filtration-hiver.local    # works
- id: 1  name: watchdog.js   running: true
- id: 2  name: pool-pump.js  running: true

$ ./myhome-local.sh shelly script list 192.168.1.21              # works, same output

$ ./myhome-local.sh shelly script list filtration-hiver          # FAILS
timeout waiting for response to method device.lookup after 14s (dst: local)
```

**So the correct instruction is: address the device by `.local` name or IP.** No daemon needs to be
started, and in particular nobody should start one as a workaround — it would publish
`myhome/energy/solar/available`, a second voice into the live pool pump's solar input.

mDNS is not uniformly reliable here, so keep the IP as a fallback: `development.local` resolved via
`ping` but not `curl` earlier today, needing `192.168.1.30` directly. `filtration-hiver` is
`192.168.1.21`.

The ergonomics problem that remains — a bare name failing with a 14-second timeout and an error
message that names neither cause nor remedy — is filed as its own issue. The blocker, as described, was
not real.

Issue #433, opened 2026-08-05
bug(pool-pump): enabling #405 solar hysteresis exhausts the Pro1 JS heap (out_of_memory)


asnowfix:
## Summary

With #405's solar hysteresis **enabled**, `internal/shelly/scripts/pool-pump.js` exhausts the JS heap
on the production Shelly Pro1 (`filtration-hiver`, `shellypro1-ec62608c0230`, fw 2.0.0) and dies with
`errors: ["out_of_memory"]`, leaving the pump uncontrolled. With solar **disabled**, the identical
script build runs stably.

This is not the #421 crash (that fix is verified working on this device — see below). It is the
solar feature's own memory cost, on a device that has almost no headroom left.

## Evidence — measured on the real Pro1, 2026-08-04/05

Total JS heap is one pool shared across all scripts (`watchdog.js` is also resident). All figures
from `Script.GetStatus`:

| script build | solar | `mem_peak` | `mem_free` | outcome |
|---|---|---|---|---|
| v0.11.9 (no solar code at all) | n/a | 17234 | 11956 | comfortable, runs all day |
| PR #426 build (#421 fix) | **disabled** | 21350 | 6398 | stable overnight |
| PR #426 build (#421 fix) | **enabled** | 22274 and climbing | — | **`out_of_memory`** |

The #421 fix itself is confirmed healthy on this device: the live debug log shows
`Tracking OK (#421): seen 33 max 4`.

False leads explicitly ruled out before concluding, via `Script.Eval` on the live device:
- water-supply protection was NOT active (`WATER_SUPPLY_ACTIVE == false`) — it can force-stop the
  pump and the production events DB shows it firing routinely, so it must be checked;
- `isMyTurnToRun()` was `true`;
- the solar payload was fresh (published 19.6 s earlier, well inside the 5 min staleness window);
- the hysteresis hold timer was correctly accumulating — the script was behaving as designed.

## What enabling solar adds

Turning on `script/pool-pump/solar-enabled` causes the script to:

1. `MQTT.subscribe("myhome/energy/solar/available", ...)` — an additional MQTT subscription and its
   handler machinery;
2. retain the `SOLAR` state object;
3. run a 30 s hysteresis tick timer;
4. `JSON.parse()` the aggregator payload on every message (~60 s cadence, matching Beem's poll).

## Suggested mitigations, cheapest first

1. **Stop publishing `sources` in the solar payload, or move it to a separate topic.**
[...]

PR #437, opened 2026-08-05
fix(pool-pump): restore pump on water-supply clear by re-deriving the schedule window (#436)

### LINKED ISSUE #436: bug(pool-pump): pump is never restored if the script restarts during water-supply protection
## Summary

If `pool-pump.js` restarts while water-supply protection is active, the pump is **never restored**
when the water supply clears. It stays off until the next schedule event — and mid-window the next
event is the *evening stop*, so the pool silently loses the rest of that day's filtration.

Found live on the production pump (`filtration-hiver`, Shelly Pro1) on 2026-08-05 while
water-supply protection was genuinely active.

## The mechanism (more subtle than "state is lost")

Boot **does** check the water supply — `continueInit()`'s "Step 2/4: Checking water supply status"
reads `input:0` and calls `handleWaterSupply(true)` if it is high. That correctly re-arms the
`WATER_SUPPLY_ACTIVE` debounce flag, so the naive "the flag is lost" failure does not happen.

The problem is *what gets saved*:

```js
if (waterSupplyActive) {
  STATE.savedOutput = STATE.activeOutput;   // <-- captured AFTER the pump is already off
  ...
  activateOutput(-1, ...);
} else {
  activateOutput(STATE.savedOutput, ...);   // <-- restores -1, i.e. a no-op
}
```

Sequence that loses filtration:

1. Pump running (`activeOutput = 0`). Water supply goes active -> protection saves
   `savedOutput = 0` and turns the pump off. Correct.
2. The script restarts for any reason (crash, re-upload, device reboot) while the supply is still
   active. `STATE.savedOutput` is **in-memory only** — `STATE_KEYS` persists `activeOutput` and
   nothing else, and the in-memory default is `-1` (`savedOutput: -1`).
3. `enforceOutputState()` mirrors the *physical* switch, which is off because of step 1, so
   `STATE.activeOutput = -1`.
4. Boot's Step 2/4 sees the supply still active and calls `handleWaterSupply(true)`, which sets
   `savedOutput = STATE.activeOutput` = **-1**.
5. Water supply clears -> `handleWaterSupply(false)` -> `activateOutput(-1)` -> **pump stays off**.

So the restore path runs, emits `pool.water_supply_restored` with `restored_output: -1`, logs
"Pump restored after water supply turned off", and does nothing. The failure is silent and looks
successful in the logs.

## Impact

The pool needs a few hours of filtration per day to stay safe. Losing a window is a real
consequence, not a cosmetic one — and nothing surfaces it: the notice event claims a restore
happened.

Water-supply protection is not rare: the production events database currently holds 24
`pool.water_supply_protected` and 14 `pool.water_supply_restored` rows.

## Suggested fix direction (implementer's judgement)

Two candidate approaches; the second is likely better:

1. **Persist `savedOutput`** alongside `activeOutput` (add it to `STATE_KEYS` / the KVS mirror), so a
   restart mid-protection recovers the pre-protection value. Simple, but it stores a value that can
   itself go stale, and it costs an extra KVS write on every protection event.
2. **Do not restore a remembered value — re-derive whether the pump should be running.** On
   water-supply-clear, ask the same question `restartCatchUp()` asks in #421/PR #426: is *now*
   inside the currently-active schedule window (and, on `main`, is solar available)? Then start or
   stay off accordingly. This is robust to restarts by construction, needs no new persisted state,
   and composes naturally with the restart-catch-up work.

Option 2 also fixes a related latent case: if the schedule window *ends* while the water supply is
active, restoring `savedOutput` would start a pump that should now be off.

**Note on ordering vs the boot check**: whatever the fix, boot's Step 2/4 must not capture
`savedOutput` from a physical state that protection has already zeroed. Either skip the save when
the pump is already off, or drop the saved-value concept entirely per option 2.

## Not affected

`doStart()` reads `input:0` live and refuses to start while the supply is active
("Water supply protection active, ignoring start request"), so `restartCatchUp()` (PR #426) cannot
defeat the protection. No change needed there.

## Test plan

- Emulator test (`internal/shelly/scripts/pool_pump_test.go`): with the pump running inside a
  schedule window, raise `input:0`, assert the pump stops; restart the script with `input:0` still
  high; lower `input:0`; assert the pump **runs again**. This test fails today.
- Add the converse: if the schedule window has ended while the supply was active, the pump must
  **not** start on clear.
- Verify `pool.water_supply_restored` reports the output that was actually applied, so the event
  cannot claim a restore that did not happen.
- `make test` must stay at its branch baseline (47 packages on `main`, 46 on `v0.11.x`).

## Affects

Both lines: `main` and `v0.11.x` (the code is equivalent in v0.11.9/v0.11.10). Fix on `main` first,
then backport — but keep the change small and local to `handleWaterSupply` so it does not conflict
badly with PR #426, which touches the same file heavily.

## Related

- #421 / PR #426 — same class of bug: after a restart the script adopts a stale view and the pump
  silently does the wrong thing. `restartCatchUp()` is the reusable "should I be running now?" logic.
- #433 — solar path currently exhausts the Pro1's heap; relevant if the fix adds runtime state.

asnowfix:
Fixes #436.

## Problem

If `pool-pump.js` restarted while water-supply protection was active, the pump was **never restored**
when the supply cleared — and the failure was silent, because the code logged "Pump restored" and
emitted `pool.water_supply_restored` as though it had worked.

Boot's Step 2/4 does re-arm the debounce flag via `handleWaterSupply(true)`, so the naive "state is
lost" failure doesn't happen. The real defect is *what gets saved*: by that point the physical switch
is already off (from the pre-restart protection) and `enforceOutputState()` has mirrored it, so
`STATE.savedOutput` is captured as `-1` — and the later `activateOutput(STATE.savedOutput)` is a
no-op.

Mid-window the next schedule event is the *evening stop*, so the pool silently loses the rest of that
day's filtration. Water-supply protection is not rare: the production events DB holds 24
`pool.water_supply_protected` and 14 `pool.water_supply_restored` rows.

## Fix — option 2 from #436: re-derive, don't remember

On water-supply-clear, instead of restoring a remembered value, ask whether the pump *should* be
running right now: `isWithinRunWindow()` reads `Schedule.List` and checks the currently-active mode's
window. The restored output becomes
`(shouldRun && isMyTurnToRun()) ? mapSpeedToSwitch(CONFIG.preferredSpeed) : -1`, and
`pool.water_supply_restored` now reports the value **actually applied** rather than a claimed one.

This is robust to restarts by construction and needs no new persisted state. It also fixes a second
latent bug: if the schedule window *ended* while the supply was active, restoring the saved value
would wrongly start a pump that should now be off — confirmed broken on `main` today.

`STATE.savedOutput` is left in place for the `pool.water_supply_protected` event; that side was never
buggy.

## Tests

Both new tests **fail against unmodified `main`**, which is what proves they cover the bug:

```
pool_pump_test.go:721: pump did not run again after water supply cleared post-restart;
                       active-output = -1 (want 0) — #436
pool_pump_test.go:822: pump started after water supply cleared outside the schedule window;
                       active-output = 0 (want -1) — #436
```

- `TestPoolPump_WaterSupplyRestoreSurvivesRestart` — the #436 reproduction.
- `TestPoolPump_WaterSupplyClearOutsideWindowStaysOff` — the converse case.
- `TestPoolPump_WaterSupplyRestoresSpeed` / `TestPoolPump_Pro1ToggleAndWaterSupply` updated for the
  new schedule-window-aware semantics.

`pkg/shelly/script`: adds an optional `DeviceState.OnEvent(name, data)` hook called from
`Shelly.emitEvent()`. Genuinely required — there was previously no way for a Go test to observe an
emitted event's payload, which test requirement 3 (assert `restored_output` matches what was applied)
depends on.

`make test`: **47 packages ok, 0 failures**, matching the `main` baseline — verified independently by
the reviewer, not only by the implementer.

## Draft — memory budget, please read before merging

Minified `pool-pump.js`: **36155 → 36962 bytes (+807)**. That is over the ~300-byte guidance in
`AGENTS.md`, and it is reported rather than hidden. The cost is the `Schedule.List` round-trip and
job-parsing loop, which is inherent to answering "should I be running now?" correctly.

Whether that is affordable differs sharply by branch:

- **`v0.11.x`** (what the production pump actually runs): ample room — `mem_peak 17234`, ~12 KB free.
  Safe to backport.
- **`main`**: the script already sits at roughly **1680 bytes** of peak headroom on the production
  Pro1, and #433 shows the solar path already exceeds the heap. +807 consumes about half of what
  remains.

Suggested sequencing on `main`: land after PR #430's top-level mangling (~6.4 KB saving), and/or
reconcile with PR #426, which introduces its own `parseHM`/window-check logic for restart catch-up.
`isWithinRunWindow()`/`parseHM()` here are deliberately named and shaped to match #426's helpers so
the two can be collapsed into one shared implementation — the diff is kept to two hunks confined to
the water-supply path specifically to avoid conflicting with #426, which rewrites much of this file.

## Related

- #421 / PR #426 — same class of bug (after a restart the script adopts a stale view and silently does
  the wrong thing), and the source of the reusable "should I be running now?" logic.
- #433 — solar path exhausting the Pro1 heap; the reason the size delta above matters.
- PR #430 — the mangling that would buy back this cost several times over.

## Commits

- fix(pool-pump): restore pump on water-supply clear by re-deriving schedule window, not a remembered value ([2971a72](https://github.com/asnowfix/home-automation/commit/2971a72975d3e75f5437afd313fd9518ef9670b6))

### DIFF
--- a/internal/shelly/scripts/pool-pump.js
+++ b/internal/shelly/scripts/pool-pump.js
@@ -1364,6 +1364,57 @@ function cycleOutputs() {
 // === WATER SUPPLY PROTECTION ===
 var WATER_SUPPLY_ACTIVE = false;  // debounce guard
 
+// Parses "M H" out of a resolved Shelly cron timespec ("0 M H * * DAYS", per
+// makeTimespec()); null for a still-symbolic spec ("@sunrise") with no
+// concrete time yet. Same shape as PR #426's restartCatchUp() helper of the
+// same name - reconcile the two when #426 lands.
+function parseHM(ts) {
+  if (!ts || ts.indexOf('@') === 0) return null;
+  var p = ts.split(' ');
+  if (p.length < 3) return null;
+  var m = Number(p[1]);
+  var h = Number(p[2]);
+  if (isNaN(h) || isNaN(m)) return null;
+  return h * 60 + m;
+}
+
+// Answers "should the pump be running now?" from the active mode's
+// start/stop schedule window (Schedule.List) instead of a remembered value -
+// correct across a restart mid-protection and once the window has ended (#436).
+function isWithinRunWindow(cb) {
+  var mode = STATE.scheduleMode || 'winter';
+  var sc = mode === 'summer' ? 'handleMorningStart()' : 'handleNightStart()';
+  var ec = mode === 'summer' ? 'handleEveningStop()' : 'handleNightStop()';
+
+  Shelly.call('Schedule.List', {}, function(result, err) {
+    if (err && false) {}
+    if (err || !result || !result.jobs) { cb(false); return; }
+
+    var startMin = null;
+    var stopMin = null;
+    for (var i = 0; i < result.jobs.length; i++) {
+      var job = result.jobs[i];
+      if (!job.enable || !job.calls || job.calls.length === 0) continue;
+      var code = job.calls[0].params && job.calls[0].params.code;
+      if (code === sc) {
+        startMin = parseHM(job.timespec);
+      } else if (code === ec) {
+        stopMin = parseHM(job.timespec);
+      }
+    }
+
+    // Unresolved (still-symbolic) timespec: skip rather than guess.
+    if (startMin === null || stopMin === null) { cb(false); return; }
+
+    var now = new Date();
+    var nowMin = now.getHours() * 60 + now.getMinutes();
+    // A start > stop pair wraps past midnight (the fixed winter 23:15 -> 00:15 run).
+    cb(startMin <= stopMin
+      ? (nowMin >= startMin && nowMin < stopMin)
+      : (nowMin >= startMin || nowMin < stopMin));
+  });
+}
+
 function handleWaterSupply(waterSupplyActive) {
   log("Water supply active signal:", waterSupplyActive);
 
@@ -1384,12 +1435,22 @@ function handleWaterSupply(waterSupplyActive) {
       log("All pumps turned off for water supply protection");
     });
   } else {
-    // Water supply is OFF (signal is LOW after invert) - restore previous state
-    log("Water supply OFF - restoring output:", STATE.savedOutput);
+    // Water supply is OFF. Do NOT blindly restore STATE.savedOutput: if the
+    // script restarted while protection was active, boot's Step 2/4 re-arms
+    // this flag from a savedOutput that enforceOutputState() already zeroed
+    // (the physical switch was off), making the restore a silent no-op
+    // (#436). Instead, re-derive whether the pump should be running right
+    // now from the active schedule window - also correct if the window
+    // ended while the supply was active, when restoring savedOutput would
+    // wrongly start the pump.
+    isWithinRunWindow(function(shouldRun) {
+      var out = (shouldRun && isMyTurnToRun()) ? mapSpeedToSwitch(CONFIG.preferredSpeed) : -1;
+      log("Water supply OFF - restoring output:", out);
 
-    Shelly.emitEvent("pool.water_supply_restored", {restored_output: STATE.savedOutput});
-    activateOutput(STATE.savedOutput, function() {
-      log("Pump restored after water supply turned off");
+      Shelly.emitEvent("pool.water_supply_restored", {restored_output: out});
+      activateOutput(out, function() {
+        log("Pump restored after water supply turned off");
+      });
     });
   }
 }
--- a/internal/shelly/scripts/pool_pump_test.go
+++ b/internal/shelly/scripts/pool_pump_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


asnowfix:
## Triage 2026-09-06 — recommend **close this PR, keep #436 open**

Reviewed as part of clearing the 11 `stale-unreviewed` PRs. This is a recommendation, not an action —
the PR is left open for the maintainer's call.

**The bug is real.** I measured the rate before judging the PR, per `CLAUDE.md`'s rule that a mitigation
must state the observed rate of the failure it mitigates. Over the full 90-day `events.db` window:
**46 `pool.water_supply_protected`, 35 `pool.water_supply_restored` — 11 unpaired**, across 10 of the 32
days that had episodes. Every unpaired day carries restart or configuration activity, matching this
issue's own premise. Full table in [#436's comment](https://github.com/asnowfix/home-automation/issues/436).

**But nothing has been unpaired since 2026-08-15**, and the campaign's reconciliation work (#524, #550,
and the init/window handling around #512) landed across that same period. Two explanations fit — it was
fixed, or no restart has coincided with an episode since — and the data cannot separate them.

**The code, however, has aged out.** This branch is **209 commits behind `main`** with **9 conflict
regions**, concentrated in `internal/shelly/scripts/pool-pump.js` and `pool_pump_test.go` — precisely
the two files campaign #401 rewrote most heavily: the reconciler, `noteRelayTransition()`,
`desiredOutput()`'s branch structure, and as of #587 the reason plumbing. A patch written against the
pre-reconciler shape does not merge into the post-reconciler shape in any meaningful sense; resolving
nine regions would mean re-deciding the change nine times without the benefit of a clean design pass.

**Recommendation:** close this PR, keep **#436 open** with the measurement attached, and re-derive the
fix against today's code **if** it still reproduces. The cheap way to find out is to read the current
init path rather than wait for another coincidence: if init already re-asserts the water-supply fact and
re-derives the window on restart, #436 closes on that evidence.

Worth checking at the same time whether **#436 and #512 are one defect** — if the window resolves to
`-2` after a restart, `desiredOutput()` has no opinion and the pump is not restored, which is #436's
symptom arriving by #512's route.

asnowfix:
## Closing — the diagnosis is kept, the patch is not

Closed on the maintainer's instruction, 2026-09-06, as part of clearing the eleven `stale-unreviewed`
PRs. **This is not a judgement that the bug is invalid** — it is a judgement that a patch written
against the pre-reconciler `pool-pump.js` is no longer the cheapest route to fixing it.

**The work is not lost.** Everything of value from this PR now lives on the issue:

- **#436 stays open** and has been refreshed with the measured failure rate:
  [comment of 2026-09-06](https://github.com/asnowfix/home-automation/issues/436#issuecomment-5557308903).
  Over the full 90-day `events.db` window — **46 `pool.water_supply_protected` vs 35
  `pool.water_supply_restored`, 11 unpaired**, across 10 of the 32 days that had episodes, every
  unpaired day carrying restart or configuration activity. The failure this PR addressed is real and
  observed, not theoretical.
- That comment also records what the data **cannot** settle: nothing has been unpaired since
  **2026-08-15**, and the campaign's reconciliation work (#524, #550, and the init/window handling
  around #512) landed across the same period. Either it is fixed, or no restart has coincided with a
  protection episode since. The table cannot separate those, and neither can grep.
- It carries the recommended next step — read the current init path rather than wait for another
  coincidence — and the open question of whether **#436 and #512 are one defect**: if the run window
  resolves to `-2` after a restart, `desiredOutput()` has no opinion and the pump is never restored,
  which is this issue's symptom arriving by #512's route.

**Why the branch specifically is not worth rebasing.** It is **209 commits behind `main`** with **9
conflict regions**, concentrated in `internal/shelly/scripts/pool-pump.js` and `pool_pump_test.go` —
the two files campaign #401 rewrote most heavily (the reconciler, `noteRelayTransition()`,
`desiredOutput()`'s branch structure, and as of #587 the reason plumbing). Resolving nine regions would
mean re-deciding the change nine times, in conflict-marker form, without the benefit of a clean design
pass against the code as it now stands.

The branch `fix/436-water-supply-restore` is left in place and can be read for its approach and its
tests (+346 lines of `pool_pump_test.go`) by whoever picks up #436.

**Follow-up: #436**, open, refreshed, and self-contained — anyone can pick it up cold from there.

Issue #450, opened 2026-08-08
bug(pool-pump): script crashes with 'Too much recursion' during water-supply protection, leaving the pump running (v0.11.11)


asnowfix:
## Summary

`pool-pump.js` (v0.11.11) died on the production pump `filtration-hiver` (Pro1, fw 2.0.0) on
2026-08-08 with:

```
Uncaught Error: Too much recursion - the stack is about to overflow
 at ...S.length;n++)if(!CALL_SLOTS[n].used)return CALL_SLOTS[n].use...
                              ^
in function "acquireCallSlot" called from ...allDone,acquireCallSlot(n,s))
                              ^
in function "call" called from ...("Switch.Set",{id:e,on:t},n)
                              ^
in function "setOutput" called from ...tput=-1,saveState(),t&&t()});return}setOutput(0,!0,function(...
```

It appears to have crashed **while water-supply protection was stopping the pump**, and the failure
left the pump **running with protection active for roughly 33 minutes**.

This is a different failure from #421 (`Too many calls in progress`); it is a **stack overflow**, not
slot exhaustion.

## Timeline, from `/var/lib/myhome/events.db` on the NAS

```
09:46:00  pool.pump_start           {"reason":"Morning start event","speed":"eco","switch_id":0}
09:46:01  switch.on
09:58:27  pool.water_supply_protected  {"saved_output":0}      <- protection triggered
   (no further script events; script found stopped with the trace above)
10:31:12  switch.off                                            <- no pool.pump_stop event
11:25:27  device.offline / 11:25:29 online
11:51:27  device.offline / 11:51:30 online
11:57:56  device.offline / 11:57:59 online
12:03:59  device.offline / 12:04:02 online
12:10:07  device.offline / 12:10:09 online
12:26:16  switch.on                                             <- external (source WS_in)
12:28:25  config_changed (manual Script.Start)
12:28:31  pool.run_window           {"start_h":8.88,"stop_h":17.12,...}
```

Two things stand out:

1. **`pool.water_supply_protected` fired at 09:58:27 but the switch did not go off until 10:31:12**,
   and that later `switch.off` has no accompanying `pool.pump_stop`, so it was not the script. The
   protection's own `activateOutput(-1)` is exactly where the stack trace points, so the most likely
   reading is that the crash happened *inside* the protection stop, leaving the pump energised.
   Water-supply protection exists to stop the pump when supply is lost; it failed to.
[...]

Issue #439, opened 2026-08-06
refactor(config): single source of truth for script config schema (JSON), replacing regex extraction and hand-maintained KVS key maps


asnowfix:
## Summary

Configuration for the Shelly device scripts is defined in **three places that must be kept in sync
by hand or by regex**, with no single source of truth:

1. **`internal/shelly/scripts/pool-pump.js`** — `CONFIG_SCHEMA`, holding `key`, `default`,
   `description` and `type` for **28** settings.
2. **`myhome/ctl/pool/pool_defaults_generated.go`** — the defaults only, produced by
   `tools/extract-pool-defaults/main.go`, which **regex-scrapes the JavaScript source**:
   ```go
   pattern := fmt.Sprintf(`%s:\s*\{[^}]*default:\s*(\d+)`, regexp.QuoteMeta(fieldName))
   ```
3. **`internal/myhome/shelly/script/pool.go`** — `PoolKVSKeys`, a **hand-maintained** map of the same
   **28** KVS key strings, complete with hand-counted comments:
   ```go
   "preferred_device_id":   "script/pool-pump/preferred",      // 30 chars ✓
   "night_run_duration_ms": "script/pool-pump/night-duration", // 32 chars ✓
   ```

The same pattern exists for **`garden.js`** (20 schema entries → `tools/extract-garden-defaults` →
`myhome/ctl/garden/garden_defaults_generated.go`) and there is a further hand-maintained key map in
`internal/myhome/shelly/script/heater.go`.

## Why this matters now

Three independent problems converge on it:

1. **Fragility.** The defaults are recovered by regex from JavaScript. `tools/extract-pool-defaults`
   has **no test at all** (`extract-garden-defaults` does). A formatting change in the JS that a
   human would consider cosmetic can silently change or break extraction. `CLAUDE.md` already warns
   that these generated files are gitignored and invisible to CI, so "a missing call produces a
   silent build failure, not a lint warning".
2. **Duplication with no enforcement.** Nothing verifies that `PoolKVSKeys`' 28 entries match
   `CONFIG_SCHEMA`'s 28. Adding a setting means editing the JS, and separately remembering to edit a
   Go map, with the KVS key string retyped by hand. The `// 30 chars ✓` comments exist because the
   Shelly KVS key length limit is being checked manually.
3. **Device heap.** `description` fields are shipped to the device and cost JS heap for no runtime
   benefit — Go never reads them (`pool_defaults_generated.go` contains only numeric constants).
[...]

PR #458, opened 2026-08-09
refactor(pool-pump): one controller, eco/day/max speeds, no election

asnowfix:
**2026-08-09 11:01 CEST** — Implements the design decision of 2026-08-09: only one controller is ever
wired to the pump, and it must not check for, coordinate with, or command another. One controller
driving another is an electrical anti-pattern.

The multi-output design stays, because a Pro3 drives three speed stages. A Pro1 is on/off, and its
single stage is `max`.

## What was removed from `pool-pump.js`

- `isMyTurnToRun()` and the preferred-device election
- `pro1On` / `pro3SwitchStates` peer tracking, and the two MQTT status subscriptions feeding them
- `turnOffPro1()` — commanding the other controller
- `safeActivatePro3()` — the "turn the other one off first" guard
- the grace-delay machinery (`startGraceDelay`, `STATE.graceTimer`) and its config key
- both cross-device branches of `handleSwitchEvent`
- `parseSwitchStatus()`, which only ever decoded peer status payloads

**336 lines out of the script.**

## This deletes the suspected cause of #450

One of those branches is the loop found on 2026-08-09: a Pro1 sees its own switch turn on, finds a
stale `pro3SwitchStates` entry, turns itself off, and re-activates after the grace delay — repeating
for as long as the stale state persists.

```js
if (anyPro3On) {
  setOutput(0, false, function() {          // off
    startGraceDelay(CONFIG.graceDelayMs, function() {
      setOutput(0, true, function() { ... });  // back on -> raises another switch event
    });
  });
}
```

It also **bypassed the anti-cycling fuse**: `fuseAllowOn()` is consulted only inside
`activateOutput()`, and this path calls `setOutput()` directly. So the one path that could cycle
indefinitely was the one the fuse could not see.

This PR removes the failure mode rather than guarding it. #450 should stay open until the pump has
run on this build, since the crash itself was never reproduced.

## Speeds are eco / day / max

Renamed consistently through the schema, the generated KVS keys, the JS and the CLI.

`high-rpm` **merges into** `max-rpm`: they were the same concept once "high" became "max", and
production carried the same value (2900) in both. `speedRpms` already aliased `max` to `highRpm`.

Config fields: **28 → 23**. `grace-delay`, `preferred`, `pro3-id` and `pro1-id` are gone.

Since #439, the JSON schema is the single source of truth, so all of this starts in
`pool-pump.schema.json` and regenerates into both the JS block and the Go constants. The
`genconfigschema` regression tests failed exactly as designed when the key set changed, and their
frozen expectations are updated here.

## CLI change (breaking)

`pool preferred <device-id> <speed>` → **`pool speed <speed>`**. There is no longer a device to
prefer. Valid speeds are `eco`, `day`, `max`; a Pro1 resolves any of them to its single stage rather
than refusing, so a device still carrying a Pro3-era `speed` is not stranded.

Say if you would rather keep `preferred` as a deprecated alias.

## Heap: measured on hardware

`mezzanine` (Pro1, fw 2.0.0), production KVS replicated, rebooted between arms, sampled until the
peak settled:

| script | `mem_peak` | free of ~23030 |
|---|---|---|
| `main` before #447 (2026-08-08) | 22778 | 252 |
| `main` + #447, solar off | 19194 | 3836 |
| `main` + #447, solar **on** | 19194 | 3836 |
| **this PR** | **16954** | **6076** |

**−2240 bytes** from this change, on top of #447's −3584. Script runs clean, no errors.

For #433 that is decisive: headroom goes from 252 bytes on Friday to **6076**, and enabling solar
costs no measurable peak (it is a config flag plus one MQTT subscription — the code is parsed either
way).

## Deployment note — needs a KVS migration

`filtration-hiver` currently carries `speed: eco` plus the now-dead `preferred`, `pro1-id`,
`pro3-id`, `grace-delay`, `mid-*` and `high-*` keys, and lacks `day-speed` / `max-speed` / `day-rpm`.
A Pro1 maps any speed to its single stage, so `speed: eco` is harmless, but the missing keys must be
written before this script runs there. The device is offline for hardware maintenance at time of
writing, which makes it a convenient moment.

## Verification

- `make test`: **47 ok, 0 failures**.
- Uploaded to `mezzanine` and run: clean init, no errors, peak as measured above.

## Commits

- refactor(pool-pump): one controller, eco/day/max speeds, no election ([b709b09](https://github.com/asnowfix/home-automation/commit/b709b09848b6abc10f3b8549b13a9f18ed11ad83))
- fix(pool-pump): a Pro1 runs at max, so the flow model must say so ([22f45c7](https://github.com/asnowfix/home-automation/commit/22f45c71044aa546221cb222c64a819a6ed00355))
- Merge main into refactor/single-controller-pump ([83ed11a](https://github.com/asnowfix/home-automation/commit/83ed11a136bc1dceafb308ea6e5401e6bcb788f9))
- Merge main into refactor/single-controller-pump ([00237f6](https://github.com/asnowfix/home-automation/commit/00237f60ab806126d08421d36f9f08b8920f4f79))
- Merge main into refactor/single-controller-pump ([f8e5887](https://github.com/asnowfix/home-automation/commit/f8e5887a3e2af08b7c1b406a927cd7d078cbd3c6))
- Merge main into refactor/single-controller-pump ([3068cfc](https://github.com/asnowfix/home-automation/commit/3068cfc55b956a708206cf3bd07ee679c2bbfd45))

### DIFF
--- a/internal/myhome/shelly/script/pool.go
+++ b/internal/myhome/shelly/script/pool.go
@@ -38,22 +38,19 @@ func NewPoolService(log logr.Logger, provider DeviceProvider) *PoolService {
 
 // SetupOptions contains configuration for pool setup
 type SetupOptions struct {
-	DeviceIDs            []string // All device IDs to set up
-	PreferredDeviceID    string   // Which device ID should run
-	PreferredSpeed       string   // "eco", "mid", "high", "max"
+	DeviceIDs            []string // Device IDs to set up
+	PreferredSpeed       string   // "eco", "day" or "max" ("max" is the only stage a Pro1 has)
 	NightRunDurationMs   int
-	GraceDelayMs         int
 	EcoSpeed             int
-	MidSpeed             int
-	HighSpeed            int
+	DaySpeed             int
+	MaxSpeed             int
 	TemperatureThreshold float64
 	PoolVolume           int
 	Turnover             int
 	MaxFlowRate          int
 	MaxRpm               int
 	EcoRpm               int
-	MidRpm               int
-	HighRpm              int
+	DayRpm               int
 	MaxTemp              float64
 
 	// Solar-driven hysteresis (#405)
@@ -70,12 +67,17 @@ type SetupOptions struct {
 	NoMinify    bool
 }
 
-// Setup configures the pool pump system on all specified devices
+// Setup configures the pool pump controller(s).
+//
+// One controller is wired to the pump at a time; several may be configured (a
+// Pro3 kept as a spare for a Pro1, say), but they never coordinate — each acts
+// solely on its own schedule. There is therefore no minimum device count and no
+// election.
 func (s *PoolService) Setup(ctx context.Context, opts SetupOptions) error {
-	s.log.Info("Setting up pool pump system", "devices", len(opts.DeviceIDs), "preferred", opts.PreferredDeviceID)
+	s.log.Info("Setting up pool pump", "devices", len(opts.DeviceIDs))
 
-	if len(opts.DeviceIDs) < 2 {
-		return fmt.Errorf("at least 2 devices required, got %d", len(opts.DeviceIDs))
+	if len(opts.DeviceIDs) < 1 {
+		return fmt.Errorf("at least 1 device required, got %d", len(opts.DeviceIDs))
 	}
 
 	// Resolve all device IDs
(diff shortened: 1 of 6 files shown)


asnowfix:
**2026-08-09 12:26 CEST** — Correctness fix pushed, from maintainer review.

My "map rather than refuse" for a Pro1's speed fixed the switching and **left the arithmetic lying**. `computeFlowRate()` scales `maxFlowRate` by `rpm / maxRpm`, so a Pro1 carrying a Pro3-era `speed: eco` modelled 2000/2900 = **69% of its real flow**. Consequences, all silent:

- turnover per day underestimated by about a third,
- `computeRunHours()` demanding a longer window than the pool needs,
- so the pump **over-filters** — and `ctl pool status` reports a turnover that never matches reality.

Fixed by `effectiveSpeed()`, which resolves the speed the device actually runs at (always `max` on a Pro1) and is now used by `mapSpeedToSwitch()`, `computeFlowRate()` **and** the `pool.pump_start` event — so switching, the flow model and the telemetry agree.

Normalising rather than rejecting is still the right call for the stranding case: a device configured in the Pro3 era keeps filtering. It just no longer lies about what it is doing.

`make test`: 47 ok, 0 failures.

Issue #476, opened 2026-08-12
refactor(pool-pump): replace the ad-hoc state machine with a level-triggered reconciler and a single serialised actuator


asnowfix:
## Context — read this first, nothing here assumes you saw any prior discussion

`internal/shelly/scripts/pool-pump.js` controls a swimming-pool filtration pump. Exactly one Shelly
device drives the pump at a time. The production device is `filtration-hiver`, a **Shelly Pro1**
(`shellypro1-ec62608c0230`); `mezzanine`, a second Pro1, is the free test bed. The script also
supports a Pro3 (three switches = three speed stages), though no Pro3 is currently wired to the
pump.

The script has grown into a state machine that was never designed as one: a set of independent
state variables correlated by conditionals spread across many call sites, mutated from several
asynchronous entry points. Four production defects trace to that shape:

- **#450** — a second `input:0` (water-supply) event arriving mid-transition re-entered
  `activateOutput` inside the first transition's `Switch.Set` completion and overflowed the
  interpreter stack (`Uncaught Error: Too much recursion`, innermost frame `acquireCallSlot`). Fixed
  by adding two more guard variables covering that one path.
- **#474** — with `solar-enabled = true`, initialisation dies inside
  `setupMQTT()` → `subscribeSolarAvailable()`. Suspected mechanism: a **retained** MQTT message
  delivered synchronously inside `MQTT.subscribe`, re-entering the decision logic before init has
  settled. Still open; an alternative reading is plain heap exhaustion.
- **#469** — daily runtime accounting reset on every restart because restore state was spread
  across loosely-typed `Script.storage` keys with no single owner. Fixed.
- **#441** — the run window was evaluated by `restartCatchUp()` and rewritten by `handleDailyCheck()`
  immediately afterwards; the pump skipped a full day of filtration. Fixed.

The study behind this issue is committed at **`docs/475-pool-pump-state-machine-study.md`** (issue
#475). It contains the full state inventory, the twelve paths that can reach the relay, the
candidate survey with measured sizes, and the reasoning for the design below. Read it before
starting — but this issue states everything needed to do the work.

## The runtime, which is the binding constraint

Shelly runs a **modified Espruino, ES5 only**. Violating any of these crashes the device or fails
silently after minification (see the "Shelly JavaScript" section of `CLAUDE.md` and
`AGENTS.md` "JavaScript Engine Limitations"):

- **~23030 bytes of JS heap, shared by every script on the device.** `pool-pump.js` peaks around
[...]

Issue #475, opened 2026-08-12
study(pool-pump): the script is an ad-hoc state machine — evaluate canonical state-machine patterns for a constrained ES5 runtime


asnowfix:
## Observation

`internal/shelly/scripts/pool-pump.js` has grown into a state machine that was never designed as
one. It is implemented with a large number of independent state variables correlated by conditional
logic scattered across many call sites, rather than with an explicit set of states and transitions.

Symptoms that follow from that shape, all observed on real hardware during the #401 campaign:

- **#450** — a second `input:0` event arriving mid-transition re-entered `setOutput` inside the
  first transition's callback and overflowed the interpreter stack. The fix added *two more* state
  variables (`WATER_SUPPLY_TRANSITION_PENDING`, `WATER_SUPPLY_PENDING_APPLY`) to guard one path.
- **#474** — with `solar-enabled = true`, initialisation dies inside
  `setupMQTT()` → `subscribeSolarAvailable()`, apparently because a retained MQTT message is
  delivered synchronously during subscribe and re-enters the decision logic before init has settled.
- **#469** — daily runtime accounting silently reset on every restart, because restore state was
  spread across loosely-typed storage keys with no single owner.
- **#441** — the run window was evaluated and rewritten by two different paths that raced.

The recurring pattern is not one bug: it is that **state changes can re-enter the decision logic from
several asynchronous entry points** (input events, MQTT messages, schedule callbacks, RPC
completions, timers), and correctness depends on humans remembering every interleaving.

## What this study should produce

**Not** an implementation. A written comparison and a recommendation, plus a follow-up issue detailing
the chosen design.

1. **Capture the current machine as it actually is.** Enumerate every state variable in
   `pool-pump.js`, every event source that can mutate it, and every transition — including the
   implicit ones. A state/transition table or diagram of the real behaviour, derived from the source,
   not from intent.
2. **Survey canonical approaches** suitable for this runtime, e.g.:
   - explicit transition table (`{state: {event: [nextState, action]}}`) driven by one dispatcher
   - hierarchical/statechart decomposition (à la Harel/SCXML/XState semantics, implemented by hand —
     **not** by importing a library)
   - single event queue with run-to-completion semantics, so no handler ever re-enters another
   - actor/mailbox style, one inbox per concern
   For each: memory cost, code size, readability, and how it prevents the re-entrancy class above.
3. **Weigh them against this environment specifically**, which is the crux:
[...]

Issue #474, opened 2026-08-12
bug(pool-pump): merged #450/#469 fixes exceed the Pro1 heap budget — script fails to initialise with solar enabled


asnowfix:
## Symptom

After the fixes for #450 (water-supply re-entrancy) and #469 (runtime persistence) landed on `main`,
`pool-pump.js` **fails to complete initialisation** on `filtration-hiver` (Shelly Pro1,
`shellypro1-ec62608c0230`) when `solar-enabled = true`. The script dies immediately and the pump does
not run at all.

Reproduced twice in succession on 2026-08-12 at 05:55 and 05:56 CEST, on `main` @ `574f52d`
(`pool-pump.js` sha256 `475086a9…`, 97939 bytes):

```
Uncaught Error: Too much recursion - the stack is about to overflow
 at ...t"?e+=JSON.stringify(n):e+=String(n)}catch(n){e+=String(argu...
                              ^
in function "log"                     called from ...ome/energy/solar/available")
                              ^
in function "subscribeSolarAvailable" called from ..."),subscribeSolarAvailable()
                              ^
in function "setupMQTT"               called from ...rceOutputState(),setupMQTT(),CONFIG.solarEnabled&&(SOLAR.ti...
```

This is **not** the #450 crash. That one arrived through the input-event path during a water-supply
flap; this one happens during `continueInit()` → `setupMQTT()`, before the script is doing any work,
and the innermost frame is `log()` performing `JSON.stringify`.

Setting `solar-enabled = false` and restarting makes the script start and run normally, which is the
state `filtration-hiver` was left in so the pool still filters.

## Measurements

All on the same build, both devices freshly restarted, `mem_*` in bytes:

| device | build | solar | `mem_used` | `mem_peak` | `mem_free` |
|---|---|---|---|---|---|
| `filtration-hiver` | new (`574f52d`) | **off** | 15806 | 21154 | **6510** |
| `mezzanine` | new (`574f52d`) | off | 15596 | 20454 | 7434 |
| `filtration-hiver` | previous | **on** | — | 21728 | 7896 |

The Pro1's shared JS heap is roughly **23030 bytes**.

So the merged fixes cost about **1.4 KB** of headroom, and the solar path (measured at ~980 bytes on
the previous build) no longer fits on top. `mezzanine` runs the identical build without trouble
because it has `solar-enabled = false` and a smaller KVS config.

"Too much recursion" here is a **symptom of memory pressure**, not of genuine call nesting: the
[...]

Issue #469, opened 2026-08-11
bug(pool-pump): daily runtime accounting resets to zero on every script restart, not just after a crash


asnowfix:
## Symptom

`pool-pump.js` loses the day's accumulated pump runtime **every time the script restarts**. The pump
then believes it has filtered far less than it has, and keeps running to make up a deficit that does
not exist.

This has been attributed to crashes (see #450), but the evidence below shows it happens on **ordinary
restarts too** — a crash is merely one way to trigger a restart.

## Evidence

`filtration-hiver` (Shelly Pro1, `shellypro1-ec62608c0230`) on 2026-08-11, six script restarts across
the day, captured over UDP. **Every single one restored zero:**

```
06:34:16  [pool-pump]  Restored runtime today: 0 s for date: 2026-8-11
06:45:35  [pool-pump]  Restored runtime today: 0 s for date: 2026-8-11
06:46:12  [pool-pump]  Restored runtime today: 0 s for date: 2026-8-11
09:54:47  [pool-pump]  Restored runtime today: 0 s for date: 2026-8-11
18:28:01  [pool-pump]  Restored runtime today: 0 s for date: 2026-8-11
19:33:23  [pool-pump]  Restored runtime today: 0 s for date: 2026-8-11
```

and three of them additionally logged a spurious day rollover, which zeroes the counter:

```
09:54:47  [pool-pump]  Runtime day rollover: null -> 2026-8-11
18:28:01  [pool-pump]  Runtime day rollover: null -> 2026-8-11
19:33:23  [pool-pump]  Runtime day rollover: null -> 2026-8-11
```

`null` as the previous date means `STATE.runtimeDate` was null at the rollover check, so
`STATE.runtimeTodaySec` was reset to 0 by the branch at `pool-pump.js:1094-1097`.

Concrete cost on the day: the pump ran **9h09m (~6.2 turnovers)**, and after the 18:26 restart the
script reported **0.54 turnovers**. Nothing would have stopped it on turnover grounds for the rest of
the evening.

## Where the state lives

Runtime accounting is persisted to **`Script.storage`**, not KVS (`pool-pump.js:1028-1037`):

```js
var savedRuntimeDate = loadStorageValue(STORAGE_KEYS.runtimeDate);
var savedRuntimeSec  = loadStorageValue(STORAGE_KEYS.runtimeSec);
STATE.runtimeDate     = (typeof savedRuntimeDate === "string") ? savedRuntimeDate : null;
[...]

PR #479, opened 2026-08-12
refactor(pool-pump): level-triggered reconciler with a single serialised actuator (#476)

asnowfix:
Implements the design from the state-machine study (#475), planned in #476.

## Why

`pool-pump.js` had become a state machine that was never designed as one: many independent state variables correlated by conditionals, mutated from several asynchronous entry points. Four production defects trace to that shape — #450 (re-entrancy overflowed the stack), #474 (init dies with solar enabled), #469 (restore state spread across untyped keys), #441 (two paths racing on the run window). Each was fixed by adding another guard variable, which made the next one likelier.

## What changes

Facts (one scalar per observed fact, one writer each) → a pure `desiredOutput()` → one `applyOutput()`/`applyDone()` pair that is the **only** code allowed to call `Switch.Set`. Every entry point becomes "write the fact, call `reconcile()`, return"; `reconcile()` sets a dirty bit and defers via `queueTask` with a **named** function, so nothing is allocated per event.

`pool-pump.js`: 852 lines changed. Migration steps 0–4 of #476, plus a step-5 optimisation removing a redundant `Schedule.List` parse at peak.

## A second, previously-unknown crash — found on the device, not in tests

During live revalidation on `mezzanine`, `enforceOutputState()` — called synchronously at the top of `finishContinueInit()`, inside the init chain #474 already established has no stack headroom — was found to call `startRuntimeAccounting()` inline whenever the output was **already ON at restart**. That reaches `log()` deep enough to overflow the interpreter stack.

Reproduced live, fixed with the same `queueTask()` deferral (`f837fb4`), re-verified on device.

**The emulator could never have caught this**: goja has no stack limit. This is the second time this class has bitten the script, so the general hazard — *init must not run deep* — is now recorded in `AGENTS.md`.

## Device validation — `mezzanine` (Pro1, script id 4)

| arm | `mem_peak` | `mem_free` | result |
|---|---|---|---|
| solar **off** | 21028 | 7182–7238 | runs |
| solar **on** | 21028 | 7182–7238 | **runs** |

Device-side code hash `18d3143399…`, 35334 bytes, fetched back off the device.

**It initialises with `solar-enabled = true`** — the first build to do so since #474 appeared. Independently re-verified by the coordinator at 14:38 CEST: `running: true`, `solar-enabled: true`, `mem_peak 21028`, `mem_free 7196`.

- ✅ `mem_peak` ≤ 21154 (acceptance)
- ✅ `mem_free` ≥ 5000 with solar on
- ❌ `mem_peak` ≤ 20000 (**target, not met**) — 21028, unchanged from before this branch. Structural, from the policy/reconciler/actuator split; not a regression introduced here.

## Also in this branch

- **`override-ms` = 7200000 (2 h)** — a manual override (button press or out-of-band switch change) holds the pump against policy for that long. Required, not optional: a reconciler otherwise reverts out-of-band relay changes. Schema + regenerated mirrors + `docs/pool-pump.md`.
- **New emulator capability**: `ScheduleEvalInjector` + `scheduleEvalForwarder`, mirroring the existing `EventInjector`, so tests can fire a Schedule job's code against a running script (the device does `Schedule.eval → Script.Eval`).
- **`TestPoolPump_WaterSupplyRestoresSpeed` had two independent bugs** in the earlier rework: a wrong switch-id assumption, and an unmocked forecast URL racing genuine network access. Note the test sandbox **does** have outbound internet — relevant to #443.

## Tests

```
go test ./internal/shelly/scripts/...
44/44 pass, 0 failures, 375.975s
```

Coordinator spot-check on the rebased branch (`WaterSupply|CallSlot|RuntimeAccounting|Reconciler|DailyCheck`): ok, 139.925s.

## Not done

- The ≤20000 `mem_peak` target.
- A "switch off at restart" device arm was not re-measured; the numbers matched the prior baseline exactly.
- **Not deployed to `filtration-hiver`.** The production pump currently runs with solar disabled. Deploying there is a separate, supervised step.

Refs #476, #475, #474, #450, #441, #469.

## Commits

- refactor(pool-pump): route every actuation through one serialised actuator (#476) ([bd5678b](https://github.com/asnowfix/home-automation/commit/bd5678b526f240a93d06212eaca413ae77c3b9e6))
- refactor(pool-pump): add facts, desiredOutput() and shadow mode (#476) ([5592a40](https://github.com/asnowfix/home-automation/commit/5592a4049bb153aa3746dfafdf70af9776e37834))
- refactor(pool-pump): switch the reconciler on and delete the superseded machinery (#476) ([627bcf2](https://github.com/asnowfix/home-automation/commit/627bcf2be12a155a31bfd7c7752224590979670e))
- perf(pool-pump): stop paying an extra Schedule.List parse at peak (#476) ([49f1285](https://github.com/asnowfix/home-automation/commit/49f1285f3de2e7fe7bc4a95ee0086f49339b02d5))
- wip(pool-pump): reconciler tests, does not compile yet (#476) ([97d16f6](https://github.com/asnowfix/home-automation/commit/97d16f62b97a378fce806d70917791580a65b1df))
- test(pool-pump): make the reconciler test suite compile (#476) ([815ead7](https://github.com/asnowfix/home-automation/commit/815ead75e7f3a19fa851e75074bb281710e7c8e4))
- fix(pool-pump): override-ms default is 7200000 (2h), not 3600000 (#476) ([00c8e99](https://github.com/asnowfix/home-automation/commit/00c8e99d1b3f4d6cc0a65bf83a1ce8b080293f93))
- fix(pool-pump): defer startRuntimeAccounting() at init, second #474-class site (#476) ([fc8280d](https://github.com/asnowfix/home-automation/commit/fc8280d81a28b300ce86377ec33fb7b548355431))
- fix(pool-pump): TestPoolPump_WaterSupplyRestoresSpeed had two bugs, not one (#476) ([0ab3fc9](https://github.com/asnowfix/home-automation/commit/0ab3fc9d930601dc6f0418c4c3afcf985b004e89))
- docs(pool-pump): record the interpreter-stack-depth-at-init hazard in AGENTS.md (#476) ([facb974](https://github.com/asnowfix/home-automation/commit/facb9746dda50fb1facdbaa863c7335e3d8f3a80))
- test(genconfigschema): register override-ms in the schema golden maps (#476) ([3e02aaf](https://github.com/asnowfix/home-automation/commit/3e02aaf3849743fdc7353638fd24bca9a50168f4))


asnowfix:
> **Provenance note (posted by the session coordinator, 2026-08-14 16:30 CEST).** The review below was
> written in full by the principal agent driving #401/#485 earlier today, and was complete on disk when
> that agent — and a sub-agent it had spawned to settle the CI failure — were both terminated by an
> account spend limit at ~10:30 CEST, before either could post anything here. It is reproduced verbatim.
>
> Two caveats on it, from me rather than its author:
>
> - **The local test run it promises in §6 never completed.** "A local run on a quiesced machine is in
>   progress to settle it and I will post the result here" refers to the sub-agent that died. Whether
>   `TestPoolPump_ButtonCyclesPro3` fails deterministically on this branch is therefore **still open**,
>   and is the first thing the next agent should settle. What *is* established in §6 — that the red
>   `coverage-delta` is a test failure rather than a coverage regression, and that neither #393 nor #274
>   grants it a flake exemption — was verified from the CI logs and stands.
> - **This PR is now `mergeable: CONFLICTING`** (checked 16:23 CEST). That makes finding 3 mandatory
>   rather than advisory: the `origin/main` merge has to happen before anything else here can be tested.
>
> No other part of the review depends on the unfinished run.

---

## Review — 2026-08-14, coordinating #401 / #485

Reviewed by fanning out to three specialists (init stack depth and correctness, silent failures and
reconciler convergence, test coverage) and re-verifying every finding below against the branch source
myself. I also ran the #486 device reproduction against the build this PR has deployed on `mezzanine`.

**Verdict: changes requested.** The central design is right and the #486 fix in it is verified on
hardware. Two defects in the new actuator need fixing before this can go near `filtration-hiver`, and
the red CI check is a real failing test rather than the flake it was assumed to be.

---

### First, what this PR gets right — including the thing the campaign most needed

**The init stack-depth fix (#486) is correct and now measured on hardware.** `finishContinueInit()`
defers with `queueTask(setupMQTT)`, and `enforceOutputState()` defers with
`queueTask(startRuntimeAccounting)` — both passing *named* functions, so no closure is allocated. I
traced the whole `init() -> loadConfig(cb) -> continueInit() -> loadState() -> finishLoadState() ->
finishContinueInit()` chain on this branch and found no other new call frame reachable synchronously
from it; the new reconciler machinery is only ever entered from event handlers, RPC completions or the
task queue.

Measured on `mezzanine` 2026-08-14 against the bytes actually on the device (35335 B, sha256
`2db28f61c0105137`), running the reproduction that kills `main` within seconds — script stopped, relay
forced **ON**, script restarted:

| t after start | running | mem_used | mem_peak | mem_free | errors |
|---|---|---|---|---|---|
| +5 s | true | 19810 | 20790 | 3220 | none |
| +15 s | true | 15792 | 20972 | 7238 | none |
| +150 s | true | 15792 | 20972 | 7238 | none |

No crash, memory flat from +15 s. Full detail on #486. This is the precondition #485 §3 required.

Also worth naming as good design, because it is the opposite of finding 1 below: `desiredOutput()`'s
`-2` "no opinion, don't guess" convention correctly distinguishes an unresolvable `Schedule.List` from
a legitimate "off", and the fuse-trip refactor returning `false` instead of calling
`turnOffAllSwitches()` inline removes a double-actuation hazard. The `RC_BUSY` / `RC_DIRTY`
re-entrancy guards and the `-2` propagation are clean.

---

### 1. BLOCKER — `applyDone()` never looks at the `Switch.Set` result, and commits intent as fact

`applyOutputOn()` does `setOutput(RC_TARGET, true, applyDone)`, and `setOutput()` forwards the callback
straight through:

```js
function setOutput(outputId, on, callback) {
  if (STATE.outputs.indexOf(outputId) === -1) {
    log("ERROR: Invalid output ID:", outputId);
    if (callback) callback();        // <-- also reaches applyDone, with no args at all
    return;
  }
  log("Setting switch", outputId, "to", on);
  Shelly.call("Switch.Set", {id: outputId, on: on}, callback);
}
```

So `applyDone` is invoked as an RPC completion with `(result, error_code, error_message)` — but it is
declared `function applyDone()` with **zero parameters** and reads none of them. Whether the call
succeeded or failed, it unconditionally does `STATE.activeOutput = RC_TARGET`, runs the
start/stop accounting, and emits `pool.pump_start` / `pool.pump_stop`.

The consequence is specific to this PR's design, and worse than the code it replaces, precisely
*because* this is now the single trusted choke point for all actuation:

- A failed **ON**: the reconciler believes the pump started. `reconcileNow()`'s
  `if (want === STATE.activeOutput) return;` then agrees with itself forever, so **there is no retry**
  — the pool does not filter for the rest of the day, and nothing reports it. The next
  `enforceOutputState()` at script restart is the only thing that re-reads real hardware.
- A failed **OFF**: the reconciler believes the pump stopped and stops enforcing, while the pump is
  actually still running.
- The invalid-output-ID path above calls back with **no arguments**, so a validation failure is
  indistinguishable from success.

A level-triggered reconciler must derive belief from observation, never from intent. Suggested shape:
check `error_code` in `applyDone`, and on failure leave `STATE.activeOutput` alone, clear `RC_BUSY`,
and re-`reconcile()` so the next tick retries — the design already gives you the retry for free once
the belief stays honest.

### 2. SHOULD-FIX — `applyDone()`'s edge detection races with `handleSwitchEvent()`

`applyDone()` recovers the pre-transition state by re-reading `var was = STATE.activeOutput` *after*
the RPC round trip. But `handleSwitchEvent()` writes that same variable unconditionally:

```js
  if (!RC_BUSY) {
    // ... override adoption is guarded ...
  }

  STATE.activeOutput = observed;   // <-- not guarded by RC_BUSY
  saveState();
  reconcile("switch event");
```

Only the override-adoption block is gated on `!RC_BUSY`; the write is not. If the relay's own
`switch:0` status-change event is delivered before the `Switch.Set` completion callback — the ordering
the project's `DeviceState.SetPendingNestedEvent` harness exists to model, built for the closely
related #450 — then `was` is already the post-transition value, both edge branches are skipped, and
**that transition's runtime accounting and its `pool.pump_start`/`pump_stop` event are silently
lost**. The settled state is still correct, which is why `waitFor(active-output)`-style assertions
cannot see it: `TestPoolPump_NestedEventDuringActuationProducesOneChain` injects almost exactly this
interleaving but asserts only final state and switch output, not that accounting and events fired.

This is the same silent-loss class as #469, and it undercuts one of this PR's own stated goals (one
actuator, therefore no missing or duplicated accounting). Fix: capture the pre-transition output at
**dispatch** time in `applyOutput()` (e.g. `RC_WAS = STATE.activeOutput`) and have `applyDone()` use
that, so delivery ordering cannot matter. Please also add an assertion on the emitted events to that
nested-event test — as written it passes either way.

### 3. SHOULD-FIX — merge `origin/main`: this branch is 6 commits behind, and those 6 are the #480 safety fixes

The review flagged that `processTaskQueue()`'s `task()` call has no `try`/`catch` on this branch, and
that `Shelly.addEventHandler`'s dispatcher is unwrapped — both serious here, because this PR routes
essentially all pump control (`reconcileNow`, `applyOutput`/`applyDone`, `readWindow`,
`startRuntimeAccounting`, `setupMQTT`) exclusively through the task queue, so one uncaught throw
anywhere in the reconciler kills the script.

Both are **already fixed on `main`** — `83fcb2e` ("wrap task-queue dispatch and script.eval payloads")
and `4ea03b7` ("wrap addEventHandler/addStatusHandler/MQTT.subscribe callbacks"), merged as #483 after
this branch was cut. So this is not new work: merge `origin/main` in. Please do it as a **merge, not a
rebase**, and expect a real conflict in `pool-pump.js` — `4ea03b7` wraps handler bodies in the same
region this PR rewrites `handleSwitchEvent`. When resolving, keep the wrapping **in place at the
callback body**; do not reintroduce it as a higher-order wrapper, since that adds a frame to every
dispatch and #486 is exactly about frames.

### 4. SHOULD-FIX — Pro3 break-before-make can leave two stages energized

`turnOffOtherOutputsNext()` ignores each step's `error_code` and always proceeds, so if one "other"
output's off-command fails silently, `applyOutputOn()` still turns the target on. Pre-existing, but
this PR makes `applyOutput()` the sole trusted actuator with nothing downstream to catch it. Related:
the failing test in the next section is the Pro3 button-cycling path.

### 5. SHOULD-FIX (needs a device check) — nothing forces a reconcile at the end of init

There is no unconditional `reconcile()` once init completes; first convergence depends on an
incidental fact write during boot (a resolvable schedule window, or water protection being active). If
neither happens, a wrong relay state inherited across a restart can go un-reconciled until some later
external event. That is #484's shape. A cheap `reconcile("init")` after `verifySchedules` completes
would close it — and it should be deferred via `queueTask` like everything else in that region.

Note this is also the likely explanation for something I observed on `mezzanine` during the #486
reproduction: the relay stayed **ON for the full 150 s** after restart, with no run window in
existence (`mezzanine` has no pump schedule jobs, so `desiredOutput()` returns `-2`). That contradicts
the intent stated in this branch's own comment, which says a restart outside the run window should
*stop* a pump left running by hand. Worth deciding explicitly which behaviour is intended and
asserting it in a test.

---

### 6. The red `coverage-delta` is a real failing test, not the #393 flake

`build` is green; `coverage-delta` fails. From the failing job itself (run `31600130314`, job
`94127405562`):

```
FAIL  github.com/asnowfix/home-automation/internal/shelly/scripts   387.807s
--- FAIL: TestPoolPump_ButtonCyclesPro3 (19.88s)
make: *** [Makefile:221: cover] Error 1
```

`make cover` exits before `PR_COVERAGE` is ever computed, so this check being red does **not** mean a
coverage regression — a test failed.

It was assumed to be the known flake, and that assumption does not hold. #393 is
*"myhome/daemon pool/solar automation tests flake under CPU contention"* — a different package. The
`internal/shelly/scripts` issue is #274, a **data race** that only manifests under `-race`, and #274
explicitly records that it does not affect CI, which runs `-covermode=atomic` without `-race`. So
there is no flake exemption for this failure. I have corrected that note on #485.

`TestPoolPump_ButtonCyclesPro3` exercises the Pro3 button-cycling path — the same path as findings 2
and 4. Two independent signals landing on one path is not a flake pattern. This PR's body claims
`44/44 pass, 0 failures, 375.975s` for that package on this branch, which contradicts CI; a local run
on a quiesced machine is in progress to settle it and I will post the result here.

---

### What I need before this is deployable

1. Findings 1 and 2 fixed — they are the same function and the same root shape (belief must come from
   observation, captured at dispatch).
2. `origin/main` merged in (finding 3), conflict resolved with wrapping kept in place.
3. `TestPoolPump_ButtonCyclesPro3` green, with an explanation of what it was actually catching.
4. Findings 4 and 5 fixed or explicitly deferred with a reason.

Then I will deploy to `filtration-hiver` myself, **outside** the day's filtration window so a failure
costs no filtration, and close #486 with the production measurement in the closing comment. #486 stays
open until then: the fix is verified on the test bed only, and PR #477 already demonstrated that a
`mezzanine` pass does not transfer to the production device.

asnowfix:
## `TestPoolPump_ButtonCyclesPro3` — measured on this branch, 2026-08-14 16:35–16:42 CEST

This settles the question the review above left open (its author was terminated before the run finished).
Both runs are on `verify-479` at **3e02aaf**, the PR head, on a quiesced machine with nothing else
running but a 2-minute device poller.

| run | command | result | wall clock |
|---|---|---|---|
| isolation | `go test -count=1 -run TestPoolPump_ButtonCyclesPro3 ./internal/shelly/scripts/` | **PASS** | 12.42 s (pkg 12.96 s) |
| CI conditions | `go test -count=1 -covermode=atomic ./internal/shelly/scripts/` | **PASS**, zero failures | **375.573 s** |

Compare CI's failing job (run `31600130314`, job `94127405562`): `--- FAIL: TestPoolPump_ButtonCyclesPro3
(19.88s)`, package `387.807s`.

**So it does not reproduce locally, and the PR body's claim of 375.975 s / 0 failures for this package is
confirmed rather than contradicted.** Note the two signals: the test took **12.42 s locally vs 19.88 s in
CI**, and the package **375.6 s locally vs 387.8 s in CI** — CI is running slower, which is what a
timing-sensitive test failing only there looks like.

What this does and does not establish:

- ✅ The red `coverage-delta` is a **test failure blocking `make cover`**, not a coverage regression —
  that part of the review stands.
- ✅ The failure is **not deterministic on this branch**.
- ❌ It is **not** covered by a known flake: #393 is `myhome/daemon`, and #274 is a data race that only
  manifests under `-race`, which CI does not use here. The review was right to reject that exemption.
- ⚠️ So this is a **previously uncharacterised timing-sensitive failure** in `internal/shelly/scripts`
  under CI load, not a resolved question. Two local passes do not prove a test is sound; they prove it is
  not reproducibly broken.

I have re-run the failed CI jobs on run `31600130314`. If they go green, the flake is confirmed on CI
itself and this stops blocking the PR — but the test's timing sensitivity should then be filed and fixed,
not left to fail intermittently. If they fail again in the same place, the failure is environmental to CI
and needs root-causing before this PR merges.

**None of this changes the review's verdict.** Findings 1 and 2 (the actuator committing intent as fact,
and the accounting race) and the mandatory `origin/main` merge are independent of the CI question, and the
branch is still `mergeable: CONFLICTING`.

asnowfix:
## Finding 5's device check is done — it reproduces at 7-hour scale, on this branch

The review above marked finding 5 ("nothing forces a reconcile at the end of init") as **SHOULD-FIX, needs
a device check**, on the strength of a 150-second observation during the #486 reproduction. I ran that check
today on `mezzanine`, which carries this branch's build.

**The relay was on for 6 h 57 min with no run window in existence and no opinion from the reconciler** —
`switch.on` at 11:13:49 (script-initiated, `source: loopback`), continuously on through 18:09 across a
2-minute poller with no gaps, cleared only by my manual `Switch.Set` at 18:11:13. Throughout, solar sat at
60 W against this device's own `solar-stop-w = 200`, and **no `pool.pump_start` or any other `pool.*` event
was emitted all day**, while `filtration-hiver` emitted its `pool.run_window` / `pool.pump_start` /
`pool.pump_stop` normally into the same database. So this was not a solar start — it is the
`enforceOutputState()`-restores-saved-output path with nothing reconciling afterwards, exactly as finding 5
predicted.

Full evidence, including the events-database timeline and what would settle the mechanism conclusively, is
on **#484**: https://github.com/asnowfix/home-automation/issues/484#issuecomment-5295572674

**This promotes finding 5 from "should-fix, unverified" to "confirmed on hardware."** It is the same defect
class as #484 in the opposite direction, and on the production pump it means a restart outside the window
leaves the pump running overnight, silently.

I am not adding it to the current developer's scope — they are working findings 1, 2 and 3, and the fix for
this one needs a decision first: *what should a restart outside any window do to a running pump?* The
branch's own comment says stop it; the code leaves it on. That decision should be made explicitly and
asserted in a test, not inferred.

asnowfix:
## Correction: the CI re-run failed again, and my "flake, not your problem" read was wrong

I said above that if the re-run came back green the flake was confirmed, and that if it failed in the same
place the failure was environmental and needed root-causing. **It failed in the same place**, so here is the
corrected position.

Re-ran the failed jobs on run `31600130314` (2026-08-14, ~14:49Z). `coverage-delta` failed again; `build` and
`build-pr-deb` passed. Same test, and the actual assertion — which nobody had pulled out of the log before —
is **not** a timeout:

```
pool_pump_test.go:457: button press 4: expected active-output=-1, got 2
--- FAIL: TestPoolPump_ButtonCyclesPro3 (19.89s)
```

So the 4th button press, which should cycle the Pro3 back to all-off, **leaves `active-output` at 2**. The
test then burns its `eventTimeout` (10 s, against `initTimeout` 25 s) waiting for a transition that never
comes, which is where the ~19.9 s comes from — the two CI runs matching to 10 ms (19.88 s / 19.89 s) is that
constant, not evidence of determinism. I initially over-read it as such.

Local attempts to reproduce, all on the PR head, all **PASS**:

| attempt | result |
|---|---|
| isolation | PASS 12.42 s |
| whole package, `-covermode=atomic` (CI's own flags) | PASS, 375.573 s, zero failures |
| isolation, `GOMAXPROCS=1` (starved scheduler) | PASS 12.91 s |

**Revised verdict: this is not a flake in the "ignore it" sense, and it is not a tight test budget either.**
It is a behavioural failure of the Pro3 button-cycling path that manifests on CI (2/2) and not on macOS/arm64
(3/3), which most likely makes it a genuine race in that path rather than a test artefact. Note this is
**the same path as review findings 2 and 4** — `applyDone()`'s edge detection racing `handleSwitchEvent()`,
and `turnOffOtherOutputsNext()` ignoring each step's `error_code` while proceeding anyway. Press 4 is
precisely the break-before-make "turn everything off" step. Three independent signals now land on one path.

Consequences for this PR:

- **`coverage-delta` stays red and keeps blocking**, since `make cover` exits before `PR_COVERAGE` is computed.
  It is still true that this is a test failure and not a coverage regression.
- Finding 2's fix (capture the pre-transition output at dispatch time, landing shortly) may well fix this too.
  **That must be checked against CI rather than assumed** — a local pass proves nothing here, as the table
  above shows.
- If it survives finding 2's fix, it needs a Linux reproduction before merge. It should not be waived.

asnowfix:
## 2026-08-14 18:58 CEST — finding 5's recommended fix is withdrawn; the maintainer has ruled on the underlying question

Finding 5 in my review above recommends *"a cheap `reconcile("init")` after `verifySchedules` completes"* to
close the gap where nothing forces a reconcile at the end of init. **Do not implement that as written.**

The maintainer has now decided the question that finding depended on — what a restart outside any run window
should do to a running pump:

> **Restore the pump's pre-restart state. Outside a window, a running pump keeps running.**

Since `desiredOutput()` returns `-2` ("no opinion") when no window exists, an unconditional init-time reconcile
risks resolving that toward "off" and stopping a pump that was deliberately running. Whatever runs at init must
**adopt** the pre-restart state rather than impose a schedule-derived opinion on it.

Two consequences for this branch:

1. The behaviour I flagged from the #486 reproduction — the relay staying **ON** for 150 s after a restart with
   no run window in existence — is **correct**, not a defect. I withdraw that part of finding 5. What is still
   wrong is the branch's own code comment asserting that a restart outside the window should *stop* a
   hand-started pump; that comment now contradicts the decision and should be corrected.
2. The invariant to hold, and to assert in a test, is: **after a script restart the pump's output equals its
   pre-restart output, regardless of window state.** That covers #484's real bug (restart inside a window
   failing to resume a running pump) and today's `mezzanine` observation (restart outside a window keeping it
   running) with one rule instead of two.

Full reasoning, the retraction of my "mirror case is a defect" claim, and an open sub-question about whether the
observed relay or the KVS-saved output is authoritative when they disagree:
https://github.com/asnowfix/home-automation/issues/484#issuecomment-5295572674 and the comment following it.

None of this affects findings 1, 2, 3 or 4, or the CI button-cycling failure.

asnowfix:
## 2026-08-14 20:52 CEST — CI is green on the merged branch. `TestPoolPump_ButtonCyclesPro3` no longer fails.

The review round landed via #490 (merged `48deb51d`), and CI on the resulting branch head:

| check | result |
|---|---|
| `build` | **pass** (8m29s) |
| `coverage-delta` | **pass** (14m16s) |
| `build-pr-deb` | **pass** (1m13s) |
| `advise`, `update-pr-description` | pass |
| `merge-policy` | fail — **advisory only**; the workflow is "Merge policy advisory" and flags large PRs for manual review before merge. Not a gate. |

**The failure that blocked this PR twice is gone.** `coverage-delta` was red solely because
`TestPoolPump_ButtonCyclesPro3` failed in CI — 2/2, always at `button press 4: expected active-output=-1,
got 2` — while passing 4/4 locally including whole-package `-covermode=atomic` and `GOMAXPROCS=1`. On the
merged branch it passes in CI.

**Attribution, stated honestly: this is consistent with the finding-2 fix but is not proof of it.** Four
changes landed together (findings 1, 2, 3 and the `onWindowJobs` containment match), and one green CI run
against a test that had failed twice is suggestive, not conclusive — the test's own history is that it passes
in most environments. The mechanism does line up: press 4 is the Pro3 break-before-make "turn everything off"
step, `active-output` is written in `applyDone()`, and both findings 1 and 2 changed exactly that function —
finding 2 by capturing the pre-transition output at dispatch (`RC_WAS`) so a `switch:0` event arriving before
the `Switch.Set` completion can no longer skip the edge branches. That is precisely the interleaving a slower
runner would expose. If certainty is wanted, reverting the `RC_WAS` line alone and watching CI would settle
it; I do not think that is worth a CI cycle.

### What remains on this PR

- **Review finding 4** — `turnOffOtherOutputsNext()` ignores each step's `error_code` and proceeds anyway, so
  a Pro3 break-before-make step can fail silently and leave two stages energized. Untouched, still open, and
  it sits on the same path as the test above. Worth fixing here given how much of this PR's argument is that
  `applyOutput()` is the single trusted actuator.
- **Review finding 5** — **withdrawn as written** and superseded by a maintainer decision: a restart restores
  the pump's pre-restart state regardless of any run window, and the observed relay is authoritative over the
  KVS-saved output. Spec at
  https://github.com/asnowfix/home-automation/issues/484#issuecomment-5296026517. That work lands after this
  PR, not in it.
- **No device deployment yet.** Nothing on this branch has run on `filtration-hiver`, and #486 stays open
  until it has: PR #477 already demonstrated that a `mezzanine` pass does not transfer to the production pump.
  When it happens it should be **outside** the day's filtration window (today's was 10:38–17:22) so a failure
  costs no filtration, and `mem_peak` must be re-measured — the production device sat at 21770 of a ~23030 B
  shared heap today, roughly 1.3 KB of headroom.

With findings 1, 2 and 3 in and CI green, my review verdict moves from **changes requested** to **approve once
finding 4 is addressed or explicitly deferred with a reason**.

asnowfix:
## 2026-08-15 12:26 CEST — correction: `TestPoolPump_ButtonCyclesPro3` is failing again. My "the blocker is gone" call was premature.

After PR #491 (finding 4) merged into this branch, `coverage-delta` is **red again**, same test, same assertion:

```
pool_pump_test.go:548: button press 4: expected active-output=-1, got 2
--- FAIL: TestPoolPump_ButtonCyclesPro3 (19.85s)
```

I wrote above that "the failure that blocked this PR twice is gone" on the strength of one green CI run after
#490. That was wrong, and I had even flagged the risk in the same comment before drawing the opposite
conclusion. The corrected tally:

| | CI | local |
|---|---|---|
| before the review round | **FAIL** ×2 (19.88 s, 19.89 s) | PASS ×3 (isolation, whole-package `-covermode=atomic`, `GOMAXPROCS=1`) |
| after #490 (findings 1–3) | **PASS** ×1 | — |
| after #491 (finding 4) | **FAIL** ×1 (19.85 s) | PASS (full `make test`, package `ok` 414.932 s) |

**3 CI failures, 1 CI pass, 5+ local passes.** So: intermittent, heavily biased toward CI, and not resolved by
the review round. One green run was luck, not a fix.

### Why this is probably not just a flaky test

Press 4 is the Pro3 **break-before-make** step, and that is precisely the path findings 1, 2 and 4 changed.
Finding 4 in particular introduced a **retry** where there was none. #493 records the interaction found while
writing finding 4's test: repeated reconciler retries each record a fuse change regardless of whether the
actuation succeeded, so ~4 retries trip the anti-cycling fuse and force everything off through
`turnOffAllSwitches()` — which has its own error-ignoring gap (#492). A retry loop that does not converge under
slower timing, or a fuse trip mid-cycle, would produce exactly this symptom.

It is also the same actuation-chain region that **crashed the production pump this morning** (#494:
`Too much recursion — the stack is about to overflow` in `acquireCallSlot`, reached from `setOutput` →
`Shelly.call("Switch.Set")` inside another output's completion callback, on the rolled-back build). Different
build, different failure mode, same neighbourhood — which is reason to treat this as a signal rather than as
CI noise.

### Status

Root-cause investigation dispatched, with an explicit instruction to distinguish **product defect** from **test
defect** and to report "undetermined" rather than guess. It has been told not to raise `eventTimeout` unless it
can show the transition genuinely completes and merely takes longer, measured.

**This PR should not merge to `main` until that comes back.** Everything else on it is done: findings 1–4
implemented, finding 5 withdrawn by maintainer decision (#484), `build` and `build-pr-deb` green, branch
`MERGEABLE` at 17 commits.

asnowfix (commented):
LGTM

